PR #21, opened 2020-03-05
wave.resource.surface_elevation: add missing sqrt in  and unit test

ryancoe:
I believe I found a missing sqrt in the `waves.resource.surface_elevation` function (the time series of the surface elevation was too small). Fixed and added unit test. Here's the ad-hoc code I used to find the problem. 

```python
from mhkit import wave
import numpy as np
import matplotlib.pylab as plt

gamma = 9
Tp = 1/3
Hs = 5e-2

f_low = 2
f_high = 4
df = 1/30

f = np.arange(f_low, f_high, df)

S = wave.resource.jonswap_spectrum(f, Tp, Hs, gamma)
S.columns = ['orig.']
axs = S.plot(style='.-')
# plt.grid(True)

dt = 0.01
t = np.arange(0,2*60+dt,0.01)
eta = wave.resource.surface_elevation(S, t)
ax = eta.plot()
ax.set_xticks(np.arange(0,t[-1]+1/df,1/df))

Sn = wave.resource.elevation_spectrum(eta, 1/dt, int(np.floor(t[-1]/df)))
Sn.columns = ['from eta']
Sn.plot(ax=axs, style='.--', grid=True)
axs.set_xlim(f_low,f_high)

plt.grid(True)

plt.show()

```

### DIFF
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -70,7 +70,7 @@ class TestResourceSpectrum(unittest.TestCase):
         S = wave.resource.jonswap_spectrum(self.f, self.Tp, self.Hs)
         eta = wave.resource.surface_elevation(S, self.t)
         dt = self.t[1] - self.t[0]
-        Sn = wave.resource.elevation_spectrum(eta, 1/dt, len(eta.values), 
+        Sn = wave.resource.elevation_spectrum(eta, 1/dt, len(eta), 
                                               detrend=False, window='boxcar',
                                               noverlap=0)
 
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -18,7 +18,9 @@ class TestResourceSpectrum(unittest.TestCase):
         self.f = omega/(2*np.pi)
         self.Hs = 2.5
         self.Tp = 8
-        self.t = np.arange(0, 60*10, 0.05)
+        df = self.f[1] - self.f[0]
+        Trep = 1/df
+        self.t = np.arange(0, Trep, 0.05)
             
     @classmethod
     def tearDownClass(self):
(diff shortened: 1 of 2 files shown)


mhkit/tests/test_wave.py
```diff
@@ -41,6 +42,25 @@ def test_bretschneider_spectrum(self):
...
+        Sn = wave.resource.elevation_spectrum(eta, 1/dt, int(np.floor(self.t[-1]/df)))
+
+        m0 = wave.resource.frequency_moment(S,0).m0.values[0]
+        m0n = wave.resource.frequency_moment(S,0).m0.values[0]
+        errorm0 = np.abs((m0 - m0n)/m0)
+
+        self.assertLess(errorm0, 0.01)
+
+        m1 = wave.resource.frequency_moment(S,1).m1.values[0]
+        m1n = wave.resource.frequency_moment(S,1).m1.values[0]
```
kaklise: It looks like m0n and m1n should be calculated with Sn.
Can we also compare the S and Sn directly?

ryancoe: Quite right! Updating this the moments, for the specific wave case being used for the test, I have to increase the error tolerance to 5%.

You could certainly also compare the spectra directly, I just wasn't sure of an elegant way to do this...

This is my first pull request... let me know how we should proceed.

ryancoe:
Per advise from @kaklise, I've fixed the bug with the spectral moments check (1959cbaf4cf76a0696ba6dee77befdedf464c3f2) and added a check on the spectral RMSE (411e6e3737109b0564aaf7dd9b4635a7ace5720d).

To do this more robustly, I edited `waves.resources.elevation_spectrum` to allow the user to (optionally) specify the overlap when call the Welch's method for spectral estimation (1959cbaf4cf76a0696ba6dee77befdedf464c3f2).

mhkit/tests/test_wave.py
```diff
@@ -43,24 +46,39 @@ def test_bretschneider_spectrum(self):
...
+        m1n = wave.resource.frequency_moment(Sn,1).m1.values[0]
         errorm1 = np.abs((m1 - m1n)/m1)
 
         self.assertLess(errorm1, 0.01)
+
+    def test_surface_elevation_rmse(self):
+        S = wave.resource.jonswap_spectrum(self.f, self.Tp, self.Hs)
+        eta = wave.resource.surface_elevation(S, self.t)
+        dt = self.t[1] - self.t[0]
+        Sn = wave.resource.elevation_spectrum(eta, 1/dt, len(eta.values),
```
ssolson: Suggest not making a call to values.
`len(eta)`

mhkit/tests/test_wave.py
```diff
@@ -43,24 +46,39 @@ def test_bretschneider_spectrum(self):
...
+        S = wave.resource.jonswap_spectrum(self.f, self.Tp, self.Hs)
+        eta = wave.resource.surface_elevation(S, self.t)
+        dt = self.t[1] - self.t[0]
+        Sn = wave.resource.elevation_spectrum(eta, 1/dt, len(eta.values), 
+                                              detrend=False, window='boxcar',
+                                              noverlap=0)
+
+        fSn = interp1d(Sn.index.values, Sn.values, axis=0)
+        rmse = (S.values - fSn(S.index.values))**2
+        rmse_sum = np.sum(rmse)
```
ssolson: This is mean square error 
`rmse = ((S.values - fSn(S.index.values))**2).sum()**0.5`

mhkit/tests/test_wave.py
ryancoe: Fixed with a02aeb4744034351b5411381e9b4352324da179f and

mhkit/tests/test_wave.py
ryancoe: Thanks @ssolson, fixed with 7ddd7de71eba408b2bc08fc83544654db9a4a475

kmruehl:
@ryancoe I know I'm really late to this thread, but thanks for using MHKiT and catching/resolving a bug!

PR #23, opened 2020-04-15
Added optional frequency bin widths to wave.resource

rpauly18:
-Added optional frequency bin width
-re-added tests for CDiP data 
-updated CDiP test JSON file with freq bin widths


rpauly18:
> This looks good. I suggest making sure some tests use the default frequency bins.

I added a test that utilizes the default frequency bins for MC data.

PR #28, opened 2020-04-29
Explicit phasing in surface_elevation (& add frequency_bins  opt)

ryancoe:
Currently, you can set the random number generator feed for `surface_elevation` via the optional argument `seed`. However, it would also be nice to explicitly set the phases to some known values. This pull request provides this functionality via an additional optional argument `phases`.

https://github.com/MHKiT-Software/MHKiT-Python/blob/362f76b09ea411aae214c80df3d90f5eb4f02471/mhkit/wave/resource.py#L186

### DIFF
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -95,10 +95,10 @@ class TestResourceSpectrum(unittest.TestCase):
         f_bins_np = np.array([np.diff(S.index)[0]]*len(S))
         f_bins_pd = pd.DataFrame(f_bins_np, index=S.index, columns=['df'])
 
-        eta_np = wave.resource.surface_elevation(S, self.t, frequency_bins=f_bins_np)
-        #import ipdb; ipdb.set_trace()
+        eta_np = wave.resource.surface_elevation(S, self.t, frequency_bins=f_bins_np)        
         eta_pd = wave.resource.surface_elevation(S, self.t, frequency_bins=f_bins_pd)
 
+        assert_frame_equal(eta0, eta_np)        
         assert_frame_equal(eta_np, eta_pd)        
 
     def test_surface_elevation_moments(self):
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -7,6 +7,9 @@ import json
 import matplotlib.pylab as plt
 import mhkit.wave as wave
 from scipy.interpolate import interp1d
+from pandas.testing import assert_frame_equal
+import inspect
+
 
 testdir = dirname(abspath(__file__))
 datadir = join(testdir, 'data')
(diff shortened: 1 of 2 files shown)


kmruehl:
@ryancoe, thanks. This is a very useful addition. I've tagged @ssolson to review the modification. My only comments are:

- change `phases` -> `phase` (to align with `seed` vs `seeds`)
- would be helpful to add an exampleor test using a pre-defined `phase`
-  @rpauly18, this needs to be added to MHKiT-MATLAB as well

rpauly18:
@ryancoe my question is will there ever be a situation where we are looking at several spectra and use different phases for each spectra? Or want to apply different phases to a single spectra to retrieve multiple elevation time series?  If so, it may be better and more consistent with other MHKiT code to have the input be a pandas DataFrame.

rpauly18:
Also, as a follow on comment for surface_elevation function, we may want to also consider adding the frequency bin sizes as an optional input to handle data from unevenly spaced frequency bins, since the bin size is used in this calculation. If you were to use CDiP spectra in this function, using f.diff() as we are now to compute delta_f would result in inaccuracies. This optional input has already been added to the wave module functions that use frequency_moment, but it should be added here as well.

ryancoe:
On using `phases` vs. `phase`: I think that `phases` is more appropriate. The argument `seed` is a single element `int`, whereas `phases` is meant to be an `np.ndarray`.

kmruehl:
@rpauly18 yes, this is a potential use case. We have this functionality in WEC-Sim, and its use case is when a user has wave data for distinct wave cases they want to reconstruct as exact time-series (e.g. they want to reconstruct distinct wave cases that were run in a wave tank, and they have the freq, S, and phase for each wave case). I think setting up a DataFrame for this would be appropriate. In WEC-Sim, we handle this by requiring the user to provide freq and S, and optionally provide phase. We then check to see if phase is provided, and generate the time-series accordingly. 

> 
> 
> @ryancoe my question is will there ever be a situation where we are looking at several spectra and use different phases for each spectra? Or want to apply different phases to a single spectra to retrieve multiple elevation time series? If so, it may be better and more consistent with other MHKiT code to have the input be a pandas DataFrame.

ryancoe:
agreed... will work on this

ssolson:
Ryan thanks for your development on MHKiT, this is great! So it sounds like there is a use for multiple phasing so we can move towards that using a DataFrame. The tests Ryan wrote are working for me. I am trying to understand the tests a little bit more. Not a change from this PR but in test_wave.py line 28 `Trep = 1/df`. Is this supposed to mean "Representative period"?

kmruehl:
@ryancoe Thanks! That's great.

ssolson:
@ryancoe will you give me access to push some changes to this branch?

rpauly18:
@ssolson @kmruehl passing the frequency bin widths is necessary if the bins are unevenly spaced. If you think about it, there is literally no way for us to determine the bin widths if they are uneven, because we do not know where the breaking point is between the given frequency values. Using "diff" we are assuming that the break point for the bins is half way between the two frequency values, but that is not always the case. This is why we were not passing the wave tests using CDiP data before I added the optional bin size input to the frequency_moment function and all functions that call on it. We were incorrectly assuming how the bins were spaced. CDiP provides their bin spacing for that reason.

ssolson:
@rpauly18 Thanks for clarifying! I will continue with the integration.

ryancoe:
the bin width issue (while certainly important) seems unrelated to the phasing functionality of this PR - right?

rpauly18:
@ryancoe Correct. I will submit another PR once this one is closed with changes adding the bin width

ssolson:
@rpauly18 I'm doing that now.

ssolson:
Apologies. I missed @ryancoe's comment. You are right, but I am basically done at this point so I am going to omnibus this PR. Sorry!

ssolson:
I changed the title so it all checks out now. :)

ssolson:
@rpauly18 would you mind reviewing this PR? @ryancoe I'm sorry for getting off-topic and scope creeping your PR. Won't happen again!

PR #29, opened 2020-04-30
Chakrabarti plot

ryancoe:
Add plot function based on Chakrabati's plot comparing relative importance of drag, inertia, and diffraction phenomena.

```python
from mhkit import wave
import numpy as np
import pandas as pd


D = np.linspace(5,15,5)
H = 8*np.ones_like(D)
L = 200*np.ones_like(D)
mvals = pd.DataFrame([H.flatten(),L.flatten(),D.flatten()],
                          index=['H','L','D']).transpose()
wave.graphics.plot_chakrabarti(mvals)
```

Screen Shot 2020-04-30 at 1 07 06 PM


mhkit/wave/graphics.py
```diff
@@ -123,3 +123,143 @@ def plot_matrix(M, xlabel='Te', ylabel='Hm0', zlabel=None, show_values=True,
...
+    inertia, and diffraction phemonena
+
+    Chakrabarti, Subrata. Handbook of Offshore Engineering (2-volume set). 
+    Elsevier, 2005.
+
+    Parameters
+    ------------
+    *args : floats or np.ndarrays or pd.DataFrame
+        If floats: H, L, D
+            where H is wave weight, L is wave length, D is characteristic length
```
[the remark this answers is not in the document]
ryancoe: **wave height**: twice the wave amplitude (what's your question?)

**wave length:** I would've liked to use `lambda`, but this is reserved in python. I could use `Lambda`, but that might be tempting fate.

**units:** everything is dimensionless, so they can be any unit of length as long as they're the same

ryancoe:
```bash
hostname:tests username$ python test_wave.py 
test_wave.py:464: FutureWarning: The pandas.datetime class is deprecated and will be removed from pandas in a future version. Import from datetime module instead.
  expected_index0 = pd.datetime(2019,8,1,0,0)
.test_wave.py:454: FutureWarning: The pandas.datetime class is deprecated and will be removed from pandas in a future version. Import from datetime module instead.
  expected_index0 = pd.datetime(2019,4,2,13,50)
.......................
----------------------------------------------------------------------
Ran 24 tests in 5.538s

OK
hostname:tests username$ ls ./
total 452
-rw-r--r--   1 username  ...    0B Mar  5 13:33 __init__.py
drwxr-xr-x  15 username  ...  480B Apr 29 13:14 data
-rw-r--r--   1 username  ...  9.9K Apr 10 08:13 test_river.py
-rw-r--r--   1 username  ...  3.3K Apr 10 08:13 test_tidal.py
-rw-r--r--   1 username  ...   64B Mar  5 13:33 test_utils.py
-rw-r--r--   1 username  ...   18K May  1 18:59 test_wave.py
-rw-r--r--   1 username  ...   36K May  1 18:59 wave_plot_chakrabarti.png
-rw-r--r--   1 username  ...   41K May  1 18:59 wave_plot_chakrabarti_np.png
-rw-r--r--   1 username  ...   41K May  1 18:59 wave_plot_chakrabarti_pd.png
-rw-r--r--   1 username  ...   63K May  1 18:59 wave_plot_elevation_timeseries.png
-rw-r--r--   1 username  ...  169K May  1 18:59 wave_plot_matrix.png
-rw-r--r--   1 username  ...   51K May  1 18:59 wave_plot_spectrum.png
hostname:tests username$ 
```

wave_plot_chakrabarti_pd.png
wave_plot_chakrabarti_pd

wave_plot_chakrabarti_np.png
wave_plot_chakrabarti_np

wave_plot_chakrabarti.png
wave_plot_chakrabarti

mhkit/wave/graphics.py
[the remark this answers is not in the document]
ssolson: it says H is wave weight :). 

Thanks for the other clarifications. I can make this change, rename L, and check the new commits Monday.

kmruehl:
@ryancoe @ssolson 

We have already used the variable `L` for "Capture Length" in [MHKiT terminology](https://mhkit-software.github.io/MHKiT/terminology.html). We do not have a variable defined for "Wave Length", but we do use `lambda` in the "Wave Number" definition for "Wave Length". This means two things need to be done:

- [ ] add `D` for "Characteristic Length" to terminology
- [ ] define a variable for "Wave Length" to terminology and update "Wave Number" definition

mhkit/wave/graphics.py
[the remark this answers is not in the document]
kmruehl: @ryancoe we defined [default units](https://mhkit-software.github.io/MHKiT/terminology.html#units) for MHKiT, do the function default should assume MKS units. However, you could add a note to the function that since the KC number is dimensionless one could define H, L, D in a different convention.

kmruehl:
@ryancoe can you allow me to make a merge commit to resolve the variable discussion?

kmruehl:
@ssolson this PR has been updated on @ryancoe's fork. I defined `lambda_w` for "Wave Length" and made some minor formatting updates for the API doc.

mhkit/wave/graphics.py
```diff
@@ -123,3 +123,143 @@ def plot_matrix(M, xlabel='Te', ylabel='Hm0', zlabel=None, show_values=True,
...
+        yy = np.pi*row['D']/row['lambda_w']
+        lab = '$H = %.2g,\\,lambda_w = %.2g,\\,D = %.2g$' % (row['H'], row['lambda_w'], row['D'])
+        ax.plot(xx, yy, 'o', label=lab)
+        # print(row['c1'], row['c2'])
+
+    if index > 0:
+        ax.legend(fontsize='xx-small', ncol=2)
+
+    ax.set_xlim((0.01, 10))
+    ax.set_ylim((0.01, 50))
```
ssolson, line 260: @ryancoe currently I can put in numbers that may be outside the range of this plot. A user may not notice unrealistic values in an array or DataFrame. 

Would the correct response be to send a warning that a value is outside of these hardcoded limits or to increase the limits of the plot?

ssolson: Example if I remove the hardcoded limits
image

ryancoe: Good point. Probably want to set the axes limits a bit differently I suppose... I guess maybe min(abs(myBounds, userVals)).

ssolson: @SeanPluemer This is the comment that needs to be addressed. Can you pull this PR and look into drawing the lines based on an arbitrary scale. My advice would be to use the current plot size as the default/ minimum and only adjust the axis if the data falls outside the range of the current x and y  limits as I show in my example screenshot above.

kmruehl:
@SeanPluemer Thank you. 

@ssolson this PR is now up to date with master, so there are many commits and file changes, but the only file modified by @SeanPluemer is `mhkit/wave/graphics.py`.

ssolson:
It looks like you have dropped the "all inertia"  label (bottom left). Please add back:

image

ssolson:
For 
D = 5
H = 8
lambda_w = .2

The lines do not reach the edges of the plot. Please fix:
image

ssolson:
@SeanPluemer  Same case as before:
D = 5
H = 8
lambda_w = .2

The "all inertia" is showing up outside the y_min. 
Could you please fix this and check a couple of other cases e.g. placing the dot North, South, East, West. Multiple points NSEW, One point North, one point south to ensure your work is catching any edge cases a user might try.

image

mhkit/wave/graphics.py
```diff
@@ -123,3 +123,169 @@ def plot_matrix(M, xlabel='Te', ylabel='Hm0', zlabel=None, show_values=True,
...
+    x = np.logspace(1,np.log10(graphScale[1]), 2)
+    y_breaking = 0.14 * np.pi / x
+    ax.plot(x, y_breaking, 'k-')
+
+    # deep water breaking limit (H/lambda_w = 0.14)
+    x = np.logspace(1,np.log10(graphScale[0]), 2)
+    y_breaking = 0.14 * np.pi / x
+    ax.plot(x, y_breaking, 'k-')
+    graphScale = list(ax.get_xlim())
+            fontstyle='italic', fontsize='small')
```
ssolson, line 244: @SeanPluemer

PR #35, opened 2020-05-20
Modifying examples to be more readable 

SeanPluemer:


examples/river_example.ipynb
```diff
@@ -340,7 +340,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Annual Energy Produced: 19.07 MW\n"
+      "Annual Energy Produced: 19.07 MWh\n"
```
ssolson, line 343: Hmm yeah looking at this again we have Energy which would be joules. This is the energy produced in 1 year though so why would this be MWh?

ssolson: So this is MW*s which to be MWh would be 19.07 MWs = 19.07/3600 MWh =0.00530 MWh = 5.30 kWh. Does that check out to you?

SeanPluemer: I am not sure where this shows that its MW*s, but based on the math shown looks right. I agree that it would be best to have it per hour, ill change this now.

ssolson: It is stated in the function `energy_produced` Returns value with units  Joules. Which if of course equivalent to a W*s.
https://mhkit-software.github.io/MHKiT/_modules/mhkit/river/resource.html#energy_produced

ssolson: If all that checks out to you then yes. Let's go ahead and change it to kWh. Make sure as you make the change that there is a sufficient explanation of the changes in the notebook. This was a good catch!

PR #38, opened 2020-06-02
Clean up of Power Module content and first tests

rpauly18:
-removed functions that will not be included in the June release
-moved functions into the characteristics submodule from power.quality and wave.performance
-added test for power.characteristics.instantaneous_frequency 
-moved tests for dc_power and ac_power_three_phase to test_power.py


rpauly18:
@ssolson and @kmruehl , I am mostly looking for your feedback on the organizational changes made in this PR. Do the submodules "quality" and "characteristics" make sense, should we rename them?

ssolson:
Rebecca this great to see. Responding only to the organizational changes I think the organization is good. 

I am not sold on the submodule name `characteristics`. I see you currently have `instantaneous_frequency`, `dc_power`, and `ac_power_three_phase` in the submodule. What other functions do you plan to add to this submodule? This currently seems to be sort of a "start here" class that allows the user to take measurements and find QoIs.

Similar questions for the `quality` submodule. Currently, the functions look exclusively at harmonics. What other functions are planned for this submodule?

If you want help developing code or looking at issues let us know

rpauly18:
@ssolson power quality has two main groups of calculations. The first is harmonics, which are included in the module currently. The second is flicker. Calculating flicker is a very long, multistep process. I have most of the code written with the exception of the last step or two (which are quite complicated). I removed the flicker related code because the first steps are kind of useless without the last. I will likely eventually combine many of the functions together for flicker as well. 

Characteristics is designed to be any and all power calculations that are not exclusively for power quality assessment. I have some code in development for RMS quantities, active and reactive power. There are many types of calculations that could be added here, and will be a constant work in progress. I am not necessarily sold on the name "characteristics", so if you or @kmruehl have any suggestions, I am open to them.

kmruehl:
@rpauly18 I'm also not sold on the `characteristics` submodule name. However, I do not have a better suggestion. Similar to @ssolson's comment, the distinction between `power.characteristics` and `power.quality` is not clear to me. While I understand the desire to remove functions that are in development and not yet ready for release, but I think having an empty `power.quality` submodule is more confusing than helpful. One suggestion would be to add private methods using the `__` syntax that, at a minimum, include function names and descriptions. That way users/developers can see what is in development. 

These comments aside, I think this is good progress and we should merge this PR into the feature branch. Once merged, we can work on the module's refinement in the feature branch.

rpauly18:
@kmruehl power.quality is not empty and does contain the harmonics calculations. I can reintroduce the other functions for the flicker calculations if you think it would be best. The main reason I removed them is because I have not yet decided which ones I want to combine, so if I put them back, I will make them hidden. I could put the functions from characteristics into quality as well, if you think that will be clearer. I was just trying to figure out a way to make it clear these functions are not just for power quality. You would use some of them for power performance assessments, for example.

PR #42, opened 2020-06-15
Loads module 

hivanov-nrel:
initial migration from fork to feature branch


ssolson:
Hey, Chris on the fatpack dependency. Can you speak to if you plan to use more functions from this package in the future? or will it most likely just be the 2 functions used here. Either way, we can move forward with the implementation/ dependency for this quarter. 

Also, could you speak to the choice of fatpack over the rainflow package? I like that the rainflow package specifically lists that it is following an ASME standard, not sure if that is true for fatpack.

hivanov-nrel:
I picked up the fatpack module because (from my very brief search awhile ago) I came across it and found that it has a lot of functionality that was well documented and straightforward to use. I have used some other functions from it, which can be useful for a more in-depth fatigue analysis, but I cant say for sure on whether we will end up utilizing them in the future.

I agree following a standard is important and one of the things that put my mind to ease when using fatpack was the text in the rainflow script which cites what the algorithm is based on (see attached image). I did check it against some validated results and it matches quite well. Nevertheless, I am open to switching to something else. It just might take some extra time to implement. 

rainflow_citation

rpauly18:
@hivanov-nrel I had to edit test_loads.py some so the tests would run. I fixed one test, but currently test_bin_stats is failing for me. The resulting data frames are not the same size. Can you take a look?

ssolson:
Guys on function get_DELs. I think the name should be changed to `damage_equivalent_load`.
Secondly, I think we should not be passing a DataFrame which then requires `chan_info` which is a tuple of a string and a slope. I think this function would take a pd.Series of the specific channel of interest and pass an m sole e.g.:
```
def damage_equivalent_load(df['Chan1'], m, binNum=100, t=600):
    """ Calculates the damage equivalent load of multiple variables
    
    Parameters: 
    -----------
    df : pd.DataFrame or pd.Series
        contains dataframe of variables/channel to calculate DEL
    
    m : lint or float?
         fatigue slope factor "m"
        
    binNum : int
        number of bins for rainflow counting method (minimum=100)
    
    t : float/int
        Used to control DEL frequency. Default for 1Hz is 600 seconds for 10min data
        
    Returns:
    -----------
    DEL : pd.DataFrame
        Damage equivalent load  
    
    """
```
A user would then need to write a loop around this function for multiple channels of interest.

rpauly18:
@ssolson I am tending to agree with you. I am starting to work on the Matlab wrappers and the chan_info variable may end up causing a problem in MATLAB, but I have not fully tested yet. It would be nice to be able to handle multiple variables at once, but it is not necessarily critical. 

@hivanov-nrel what are your thoughts?

hivanov-nrel:
Yeah I originally had it coded that way, but added the multiple channel functionality as I personally found it very useful. Like you said, it is not necessary, but now it puts more pressure on the user to find a way to loop through all the variables and apply the correct varying fatigue slope factor. As for the naming, it may be good to also add "short_term" in there as well because I think eventually I will write a lifetime damage equivalent load function. Let me know if you want me to apply the updates.

PR #46, opened 2020-06-26
Feature loads

kmruehl:
Updated feature_loads with the following:
- updated the example for clarity and consistency with other examples
- updated formatting for api, e.g. parameters and returns
- updated function names to be consistent with existing methods (e.g. `matlab2datetime` vs `matlab_to_datetime`)
- updated variable names for clarity and consistency (e.g. `get_stats` vs `plot_statistics`)

Comment/Question:
- [x] `loads.bin_statistics`
   - [x] Add IEC TS to function definition
- [x] `loads.damage_equivalent_load`
   - [x] This should be updated so that I can calculate the DEL on data multiple signals at once.
   - [x] This function needs a reference. Is this an IEC TS? Where does the user find the theory/equation behind this method?
   - [x] add rainflow counting algorithm to description @hivanov-nrel 
- [x] `utils.get_statistics`
   - [x] What is the added value of `utils.get_statistics` over using the native pandas functionality of: `df.mean()`, `df.min()`,`df.max()`, and `df.std()`? @hivanov-nrel
   - [x] Refers to a minimum amount of data required to run this method. What is the basis of this minimum amount of data? Is there an IEC TS?    @hivanov-nrel 
   - [x] Why is the return indexed by the first timestamp of the corresponding statistical window?
   - [x] update function description to explain that this function applies IEC standard (and how it differs from builtin functions) 
@hivanov-nrel 
- [x] MHKiT Loads Example
   - [x] In the example there is a warning, "Warning: some bins may be empty!", is that normal? Can we apply QC to this data?
   - [x] in example `[bin_DELs, bin_DELs_std] = loads.get_statistics(maxs,compare_against,bin_edges)` is called, but should it be applied to DELs? @hivanov-nrel
   - [x]  It would be useful for the example to demonstrate how to do this: "At this point, it would be nice to start visualizing some of this data. In order to do this, we need to calculate the stats and DELs for all the files in our database. In this case, it would be done through a loop that imports each file and applies all the functions we just saw. At the end of the loop, we store the result by appending to a list (this is computationally more efficient than concatenating dataframes). Finally, we can convert our lists to dataframes so that its easier to play with the data. To speed things up, this was already done so we just need to import the resulting dataframes."  @hivanov-nrel


rpauly18:
@kmruehl
loads.damage_equivalent_load
This should be updated so that I can calculate the DEL on data multiple signals at once.

We actually asked Chis to remove this functionality because the methodology to do so using the data frame column names became cumbersome. We should readdress this for a later version, but for now, we opted to keep it as working on a single signal.

rpauly18:
It would be useful for the example to demonstrate how to do this: "At this point, it would be nice to start visualizing some of this data. In order to do this, we need to calculate the stats and DELs for all the files in our database. In this case, it would be done through a loop that imports each file and applies all the functions we just saw. At the end of the loop, we store the result by appending to a list (this is computationally more efficient than concatenating dataframes). Finally, we can convert our lists to dataframes so that its easier to play with the data. To speed things up, this was already done so we just need to import the resulting dataframes."

@kmruehl are you suggesting we show a block of code that demonstrates this but is commented out, like we did for requesting data in the tidal module? We do not want to actually run that code in the example in an effort to cut back on the amount of data files we need to store.

kmruehl:
> 
> 
> @kmruehl
> loads.damage_equivalent_load
> This should be updated so that I can calculate the DEL on data multiple signals at once.
> 
> We actually asked Chis to remove this functionality because the methodology to do so using the data frame column names became cumbersome. We should readdress this for a later version, but for now, we opted to keep it as working on a single signal.

kmruehl:
> 
> @kmruehl are you suggesting we show a block of code that demonstrates this but is commented out, like we did for requesting data in the tidal module? We do not want to actually run that code in the example in an effort to cut back on the amount of data files we need to store.

That would be a good way to handle it. I like the idea of providing code demonstrating how to apply this methodology, even if it is not used in the example. However, since the necessary data files are not provided, we should not allow users to uncomment it. I think we should just provide it as inline code in the text section of the example (e.g. inline code using Markdown).

PR #47, opened 2020-06-26
fixing wave notebook

rpauly18:
Reverted wave notebook to same commit as master


rpauly18:
@kmruehl This is just a PR to fix the issue with the wave_example notebook.

kmruehl:
@rpauly18 this is one way to resolve the issue, and will fix the broken `wave_example.ipynb` file on this branch. 

I would actually recommend removing tracking of the unrelated files from feature branches altogether. That way when we merge into master, only new features are added, and no unrelated files are changed. The way to do this is to `git add` only the files relevant to the feature branch.

PR #50, opened 2020-06-30
converted pd.datetime() to just datetimeI()

SeanPluemer:


kmruehl:
@rpauly18  and @ssolson my only comment about this PR is that we are not consistent with the way we `import datetime` into MHKiT. I think we need to specify a standard approach, like we do for `import numpy as np` and `import pandas as pd`. 

Here are the other instances of `import datetime` in other MHKiT modules:
https://github.com/MHKiT-Software/MHKiT-Python/search?q=%22import+datetime%22&unscoped_q=%22import+datetime%22

ssolson:
The changes in `test_wave.py` from pd.datetime to using the datetime module has been incorporated into #53.

PR #62, opened 2020-08-19
WDRT: environmental_contours

ssolson:
This function allows users to draw environmental contours using the IFORM and modified principal component analysis as described in the [WDRT](https://github.com/WEC-Sim/WDRT) as shown in the plot below. These functions are a subset of the [environmental sampling example](https://github.com/WEC-Sim/WDRT/blob/master/examples/example_envSampling.py) which were located in the [ESSC.py](https://github.com/WEC-Sim/WDRT/blob/master/WDRT/ESSC.py) file.

This PR primarily adds one new user function and an example jupyter notebook:
 - `wave.resource.environmental_contour`
 - `examples.environmental_contour.ipynb`

image

Additional Changes:
1. Simplified the Travis build file and added a Python 3.8 Test
2. Made the year returned from available data an integer (and therefore non-unique but easier to select data)
3. Added a `wave.io.ndbc.to_datetime_index` function which sets a DateTime index and removes the NDBC date/ time column information

[Build Passing](https://travis-ci.org/github/MHKiT-Software/MHKiT-Python/builds/723932983)
[Build Status](https://travis-ci.org/MHKiT-Software/MHKiT-Python)


rpauly18:
@ssolson my initial comment for these functions is that plot_environmental_contour needs the header documentation updated to reflect the function's purpose, input parameters and output.

ssolson:
...and add a test for it? I forgot about this function apologies let me fix that.

mhkit/wave/resource.py
```diff
@@ -684,3 +686,259 @@ def func(kk):
...
+    x1: array like
+        Component 1 data
+    x2: array like
+        Component 2 data        	
+    dt : float
+        x1 and x2 sample rate (seconds) 
+    period : np.array
+        Desired return period (years) for calculation of environmental
+        contour, can be a scalar or a vector. If vector, must be an n by 1 array. 
+    PCA: dict (optional)
```
rpauly18, line 722: Sterling, I do not fully understand the scope of the PCA variable. Is this in case the user wants to re-use the PCA from a previous time running this code? Or can they compute the output from PCA using their own code/ methods? Please make this doctoring more descriptive. Also, we should not reference a hidden function in the doctoring since the hidden function will not appear in the MHKiT documentation.

mhkit/wave/resource.py
```diff
@@ -684,3 +686,259 @@ def func(kk):
...
+	    properties returned from _principal_component_analysis
+    size_bin : float (optional)
+        Data points in each bin 		
+    nb_steps : int (optional)
+        Discretization of the circle in the normal space used for
+        IFORM calculation.
+	return_PCA: boolean (optional)
```
rpauly18, line 729: Similar comment here, we should not reference a hidden function.

mhkit/wave/resource.py
```diff
@@ -684,3 +686,259 @@ def func(kk):
...
+
+    Returns
+    -------
+    x1_Return : np.array
+        Calculated x1 values along the contour boundary following
+        return to original input orientation.
+    x2_Return : np.array
+       Calculated x1 values along the contour boundary following
+        return to original input orientation.
+    PCA: Dictionary (optional)
```
rpauly18, line 741: It appears there are more keys that get returned than the ones that are listed here, Please add definitions for all of them.

mhkit/wave/graphics.py
```diff
@@ -294,3 +294,49 @@ def plot_chakrabarti(H, lambda_w, D, ax=None):
...
+
+    #assert isinstance(eta, pd.DataFrame), 'eta must be of type pd.DataFrame'
+    
+
+    ax = _xy_plot(x2_contour, x1_contour, 'k-', 
+                  label=contour_label, ax=ax)
+    ax = plt.plot(x2, x1, 'bo', alpha=0.1, 
+                  label=data_label) 
+    plt.legend(loc='lower right')
+    plt.xlabel('Energy period, $T_e$ [s]')
```
rpauly18, line 339: The input variables are generic for this function, yet the x and y labels are explicit. Will users use this function for parameters other than Hm0 and Te? If so, we should make the x and y labels optional input. If not, we should have more explicit input parameter names.

mhkit/wave/resource.py
ssolson: No these are all there are see lines 938-942

mhkit/wave/graphics.py
```diff
@@ -294,3 +294,49 @@ def plot_chakrabarti(H, lambda_w, D, ax=None):
     ax.set_ylabel('KC parameter, $\\frac{H}{D}$')
 
     plt.tight_layout()
+
+def plot_environmental_contour(x1, x2, x1_contour, x2_contour,
```
rpauly18, line 298: I would like to see this function be b able to plot multiple contours for different years simultaneously (using the direct output from environmental_contour) with each contour a different color, and with each contour having a different label. I am not sure if that change would be best made here or in the hidden function located in river.graphics that this is calling.

mhkit/wave/graphics.py
ssolson: Yeah I'm pretty torn on this. One of the major questions on WDRT is "can I use variable "x" to ..." and the answer is always "It is a statistical method, you can pass any variables you want".  I guess we can use a nondescipt "T" for the x-axis and Hs for the y-axis. lmk if you have thoughts on how to better generalize this

mhkit/wave/graphics.py
ssolson: Feel free to add that functionality.

ssolson: Wrapping multiple calls is something I would add in the next stage where we start bootstrapping the contours. See the WDRT environmental sampling example for a visual of this. This is also where the full functionality of passing a PCA becomes useful.

mhkit/wave/resource.py
ssolson: Agreed I will remove.

mhkit/wave/resource.py
ssolson: Passing the PCA variable allows you to skip calculating a PCA for the passed x1, x2. In the next phase of this PR we will be sampling the environmental contours function this function will use the same PCA each time.

mhkit/wave/resource.py
rpauly18: Ahh, I see now that sigma_param is a dictionary. It was displaying funny on my screen making it look like it's keys were keys of PCA.

rpauly18 (approved):

mhkit/wave/resource.py
kmruehl: I know this is an outdated comment, but I shared @rpauly18's confusion with providing a PCA and made minor updates to the PCA description based on this discussion.

mhkit/wave/graphics.py
kmruehl: @rpauly18 @ssolson I personally think it should be generalized to develop a contour for any two provided variables. For WECs the obvious use is T vs Hs, but what if you wanted to know Tp vs wind_speed, or Hs vs current, these could be applicable to other MRE applications and offshore wind.

kmruehl:
@ssolson this is a draft

- [x]  define acronyms prior to use, e.g. PCA and CDF
- [ ]  what does "array like" mean, e.g. "x1: array like"? I don't think we use this elsewhere in MHKiT, should this be a numpy array?
- [x]  "for the passe x1, and x2" passed?
- [ ]  would prefer "bin_size" (instead of "size_bin") for consistency with the loads module. 
- [ ] `plot_environmental_contour` does H have to be Hs? I can't think of why it would be anything else, like T can be Tp or Te. Also, I think this should technically be Hm0, not Hs. 
- [x] `plot_environmental_contour` add units

mhkit/wave/resource.py
```diff
@@ -684,3 +686,278 @@ def func(kk):
...
+        call. When supplied, the function will skip the PCA calculation 
+        for the passed x1 and x2, and instead use the provided PCA.
+    size_bin : int (optional)
+        Data points in each bin 		
+    nb_steps : int (optional)
+        Discretization of the circle in the normal space used for
+        I-FORM calculation.
+	return_PCA: boolean
+	    Default False, if True will retun the PCA dictionary 
+
```
kmruehl, line 735: Per previous comment, I think the default should be to return `PCA` if it has been calculated, and not provided. This means I think the `return_PCA` parameter should be removed.

ssolson:
- [Array-like](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects#:~:text=The%20term%20%22array%2Dlike%22,of%20the%20array()%20function.) - Any sequence that can be interpreted as an ndarray. This includes nested lists, tuples, scalars and existing arrays.
     - Array-like is currently used throughout the tidal and river modules
    - These should all be named `array-like` for consistency 

 - Let's change it to `bin_size` then

 - `plot_environmental_contours` can be any variables. I changed them to be specific per Rebecca's feedback. If someone wanted to they could use peak H, mean H, H_{1/10} ect.

- The only use of the `_principal_component_analysis` function is by/ with the `environmental_contour` function. The PCA dict is a means to an end and I believe the general user would prefer to skip this step. Further, the PCA function is a highly specialized function that takes the same inputs as  `environmental_contour`. You would never use `_principal_component_analysis` for anything else. 
    - PCA is passed to skip the calculation of the PCA (again this becomes more obviously useful in the next/ last step of the WDRT environmental sampling example being added to MHKiT)
    - return_PCA is an option because it is a more advanced usage of the function and not expected by a novice user (e.g. if I asked for the environmental contour why did I get a PCA dict?). Forcing the user to request it means they are expecting it and therefore probably know how to use it.
    - return_PCA is returned so that the PCA may be passed back to the function to skip


image
image

kmruehl:
responding to @ssolson 
> 
>     * [Array-like](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects#:~:text=The%20term%20%22array%2Dlike%22,of%20the%20array()%20function.) - Any sequence that can be interpreted as an ndarray. This includes nested lists, tuples, scalars and existing arrays.
>       
>       * Array-like is currently used throughout the tidal and river modules
>       * These should all be named `array-like` for consistency
Thanks for the description, I was not aware it is used elsewhere, I marked this as resolved

>     * Let's change it to `bin_size` then
Do you want me to make this change, or will you?
> 
>     * `plot_environmental_contours` can be any variables. I changed them to be specific per Rebecca's feedback. If someone wanted to they could use peak H, mean H, H_{1/10} ect.
I understood @rpauly18's comment differently, because it related to the plotting script. I understood her comment to mean we should either explicitly state that the function is used only for T and H, or we should generalize. I'm in favor of generalizing. 
> 
>     * The only use of the `_principal_component_analysis` function is by/ with the `environmental_contour` function. The PCA dict is a means to an end and I believe the general user would prefer to skip this step. Further, the PCA function is a highly specialized function that takes the same inputs as  `environmental_contour`. You would never use `_principal_component_analysis` for anything else.
>       
>       * PCA is passed to skip the calculation of the PCA (again this becomes more obviously useful in the next/ last step of the WDRT environmental sampling example being added to MHKiT)
>       * return_PCA is an option because it is a more advanced usage of the function and not expected by a novice user (e.g. if I asked for the environmental contour why did I get a PCA dict?). Forcing the user to request it means they are expecting it and therefore probably know how to use it.
>       * return_PCA is returned so that the PCA may be passed back to the function to skip
> 
I see your point, that the PCA is a very specialized construct, but since it's the basis of `environmental_contour` I'm still in favor of returning it _if_ it's calculated and not provided by the user. What I mean is that _if_ PCA=None, then the function should return the calculated PCA, _if_ PCA is a given parameter, it should not be returned.

ssolson:
I can change it to bin size

I believe were are agreed that PCA is a private function now? So is that resolved?

I strongly disagree with having an automatic and dynamic number of output variables. Changing the number of outputs a function returns without asking the function to change is extremely confusing in my opinion. I believe the function should always return the same outputs unless I specifically ask it to do otherwise. I would suggest we make PCA a non-private function over an automatically dynamic number of output variables but as I have said PCA is a highly specialized function that will only be used with environmental contours. So while I would disagree with making it non-private I would disagree less than the automatic dynamic output variables.

I do not understand what you are suggesting wrt to the plot. By generalize do you mean change it from Hs to H? Or do you mean to take it all the way back to my original x1, x2 docstring and change the axis to read x1 and x2? If so one thing I think is confusing from a general point of view is that while I typically think in terms of Hs and Tp analogous to x and y (e.g. in that order), for the contour plots Tp is on the x-axis. Which maybe I'm the only one who thinks this is confusing.

rpauly18:
@ssolson I believe @kmruehl is suggesting we generalize the plotting function to take any input data. To do so, we would need to make the x and y axis labels input options so that is does not default to Te and Hs. If there are going to be users who will want to use these functions for other parameters, we should generalize it.

rpauly18:
For the return_PGA discussion, How about we always return it, even if it was provided? The drawback would be taking us more memory space with a variable, but it would help the user keep related info together. Just spitballing to come to a compromise here.

ssolson:
Thanks for the suggestion, Rebecca. I would not be in favor of always returning it. The function is called environmental_contours and when I call it I expect it to return just that... the environmental contours.  I think always returning the PCA only adds confusion because now a variable that the user did not know they needed or ask for is being returned. Returning only the x1_contour and x2_contour is the most straightforward and simple return.  Returning the PCA is advanced functionality for someone looking to optimize performance on multiple env_contour calls. I think providing this as an option to return will allow the advanced users to find this functionality as they begin to work with the code more.  Additionally, if you always return it you must always set the output to be 3 variables e.g. the user must always specify x1_con, x2,_con, PCA = env_cont... I think this is really clunky and I would not look forward to this workflow as I move into the next stages of the env_contours work with the bootstrapping methods. 

The motivation for the setup I have given is that a typical user who wants to calculate an environmental contour will never intuitively know they need to perform a PCA with 2  regression fits first (to be clear here this is not strictly a PCA it is a very specific PCA developed for the WDRT). I believe that since the PCA inputs are a subset of environmental_contour why make the distinction (e.g. why make it a separate function)? The issue then is that in order to be computationally efficient we need to return the PCA as to not unnecessarily perform the PCA continually. Enter the optional PCA return solution.

There are countless functions we use throughout MHKiT (in numpy, pandas, etc) with options none of us use, know, or understand but we are still capable of using the core function (e.g. [pd.read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html)). The inputs we are discussing are clearly marked as optional and I am all for making them more clear on how to use. However, I believe the solutions you guys have provided (of which I had both implemented at one time during the draft phase) are not as clean and do not address all the issues I point out above. I agree this is not perfect a perfect solution but I believe it is the best solution on the table.

rpauly18:
Sterling, your rationale makes sense to me, and I do not have an issue making it an optional output. As you said, many users may end up just ignoring it, but it is nice to have for the more "advanced" user. As long as we have good descriptions in the docString that call out the purpose, it should be fine.

kmruehl:
I appreciate the back/forth discussion of `environmental_contour`, the role of `PCA` and `return_PCA`, and agree with the proposed solution to leave it as is. 

The reason I mentioned this is the first place was not actually because I wanted/needed `PCA` to be returned, but rather that I found it confusing that the `environmental_contour` function parameters included both the option to provide PCA as in input, and return it as an output. My thought was, why would I provide something as an input and also want it as an output? It just didn't make sense to me. This discussion clarified the intent, and I think it would be helpful to provide more clarification in the on the role of `PCA` versus `return_PCA` in the `environmental_contour` function parameters description.

kmruehl:
> 
> 
> @ssolson I believe @kmruehl is suggesting we generalize the plotting function to take any input data. To do so, we would need to make the x and y axis labels input options so that is does not default to Te and Hs. If there are going to be users who will want to use these functions for other parameters, we should generalize it. My original comment was that we needed to choose to go one way or the other. Either have the input variables be agnostic (as you originally had it) and remove the axis labels, or prescribe the input variables and keep the axis labels (as it is right now).

Yes, this was my intent. We should either say that `environmental_contour` is just used for T and H, and thus `plot_environmental_contour` plots T and H, **OR** we should generalize `environmental_contour` and `plot_environmental_contour` for any two environmental parameters. My preference is to generalize it for the reasons I mentioned earlier (e.g. what if you wanted to know Tp vs wind_speed, or Hs vs current, these could be applicable to other MRE applications and offshore wind). 

Right now there seems to be a little bit of both, where `environmental_contour` is generalized for `x1` and `x2`, but `plot_environmental_contour` is for `Hm0` and `T`.

ssolson:
Thanks Kelly! I took a stab at resolving the open issues from the above discussion this morning. 

To make the plot more general with the x_label and y_label I made all the optional inputs **kwargs. 
Then In addition to trying to clarify the PCA I made the optional inputs **kwargs there as well. 

Lastly I updated my notebook example to account for the changes. @rpauly18 & @kmruehl have you guys reviewed the notebook example? I would be surprised if I got it perfect on my first try 😂!

kmruehl:
> 
> 
> Thanks Kelly! I took a stab at resolving the open issues from the above discussion this morning.
> 
> To make the plot more general with the x_label and y_label I made all the optional inputs **kwargs.
> Then In addition to trying to clarify the PCA I made the optional inputs **kwargs there as well.
> 
> Lastly I updated my notebook example to account for the changes. @rpauly18 & @kmruehl have you guys reviewed the notebook example? I would be surprised if I got it perfect on my first try 😂!

Thanks @ssolson. I reviewed the notebook previously and made minor edits, but I'll take a look at both the code and notebookd revisions right now.

kmruehl (approved):
Okay, I just did a final review, resolved some outstanding issues with e1940ca, and I think this PR is ready for a merge. @ssolson Thanks for your hard work on this!

rpauly18:
@ssolson The example notebook and changes you made following Kelley's review look good to me. Great job on this!

PR #65, opened 2020-08-26
Download WW3 global model data

ryancoe:
I recently wrote some code to pull data for [NOAA's global WW3 model](https://pae-paha.pacioos.hawaii.edu/erddap/griddap/ww3_global.html).

Example: 
```python
import mhkit.wave.io.ww3 as mio
import datetime
import xarray as xr
import cartopy.crs as ccrs


date = datetime.datetime(2011,1,1)
fn = mio.request_data(date,23,['Tper'])
# fn = '2011-01-01.nc'
ds = xr.open_dataset(fn).squeeze().drop('depth')
fig = plt.figure(figsize=(16/2,10/1.5/2))
ax = plt.axes(projection=ccrs.Orthographic(-80, 35))
p = ds['Tper'].mean('time').plot.contourf(x='longitude',
                                          y='latitude',
                                          ax=ax) 
plt.title('Average wave period on {}'.format(date.strftime('%Y-%m-%d')))
plt.show()
```
image

You will probably want to add:

- [ ] support for firewalls (I have done this before in the past with NDBC, but for whatever reason I couldn't get it working here...)
- [ ] unit tests
- [ ] check the what the most recent data available is
- [ ] better passing of errors from website, for example when the file requested is too large (must be less than 2 GB)

### DIFF
--- a/mhkit/wave/io/ww3.py
+++ b/mhkit/wave/io/ww3.py
@@ -162,8 +162,8 @@ def request_data(start_date, hours, varnames, latBounds=None, lonBounds=None, sa
     
     Returns
     ---------
-    outUrl: string
-        URL for download
+    ofn: string
+        Saved data file name
     """
     assert isinstance(start_date, datetime.datetime), 'start_date must be of type datetime.datetime'
     assert isinstance(hours, (int, float)), 'hours must be of type float'
@@ -194,3 +194,5 @@ def request_data(start_date, hours, varnames, latBounds=None, lonBounds=None, sa
     outname = '{fn}.nc'.format(fn=start_date.strftime('%Y-%m-%d'))
     ofn = os.path.join(savepath, outname)
     _downloadFile(url, ofn)
+
+    return ofn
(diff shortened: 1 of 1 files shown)


ssolson:
Hey, Ryan thanks for submitting this feature. When you get a chance could you provide an example in your introduction to the PR? On the proxy issue, the latest version of the requests module expects the proxies to be passed as a dictionary e.g. proxy={'http': http://www.yourProxy.com/80, https: https://www.yourProxy.com/80}.

ryancoe:
@ssolson - On the proxy issue: I am actually using `urllib`, not `requests` (I've updated the imports now w. 
1d259be). I actually messed around with this for about an hour, and used all of the various libraries that can do this. I can get it to work if I [ignore the SSL certificate](https://stackoverflow.com/questions/10667960/python-requests-throwing-sslerror), but apparently this is dangerous... Certainly someone more knowledgeable can probably get this to work, but that's not me.

ssolson:
Close in favor of #233

PR #67, opened 2020-08-28
Ndbc bug

rpauly18:
Fixing a bug by allowing a single filename to be passed into ndbc.request_data as a DataFrame.


ssolson:
Rebecca is this for the MATLAB side so that it can pass filenames as a DataFrame? .. And the bug was that after the squeeze for a MATLAB DataFrame we needed to convert to a series?

rpauly18:
@ssolson it is primarily for the sake of the MATLAB side, but the bug would effect someone passing the DataFrame from Python as well. It turns out that if you have a DataFrame with a single string in it (in this case a single file name) .squeeze() will return a string, not a Series. So, converting to Series after converting to a string was necessary. I know it is a bit goofy, but this was the only combination of type conversions I could get to work.

ssolson:
Okay, the only reason a DataFrame is allowed to pass to the function though is because of MATLAB. I think the docstring needs to say series or `filenames: Series or DataFrame` instead of only `DataFrame`. This looks fine but could you add a test for this case?

rpauly18:
yes, I will edit the docstring and add a test to this PR.

ssolson:
Hey, Rebecca playing around with this function I saw that `request_data` will fail if you pass an empty Series. We need to add an assert statement that the passed series must be non-empty/ of length 1.

I was able to fix this by adding thollowing assert: 
`assert len(filenames)>0, "At least 1 filename must be passed"`
and I test that it was caught with the following test:
`ndbc_data = ndbc.request_data('swden', pd.Series(dtype=float))`

If you can add this while you're there that would be great otherwise I can add it after you update.

rpauly18:
@ssolson will do.

PR #70, opened 2020-09-03
Ndbc checks for errors while requesting files 

rpauly18:
This PR adds exceptions that catch the known possible errors when requesting NDBC data. 

- The first check is for issues decompressing files. A warning message comes up instructing the user to re-run their program to acquire the data. The program will keep running after throwing the message. Since this is a sporadic error, no unittest could be written for this code. 

- The second check is for missing data in a requested file. Again, an error message is printed to the screen and the program will keep running requesting data from other files. A unittest was written for this code using a known empty file. If NDBC ever replaces the file, the unittest could fail. 

In test_wave.py I had to add the imports: 
import contextlib
from io import StringIO

I do not believe the addition of required submodules will impact the build, but we should double check.


ssolson (commented):
Hey Rebecca this looks really good. The only thing I want to take a closer look at is the context lib part in `test_wave` which I will complete tomorrow. Thanks for taking care of this!

ssolson:
Hey, Rebecca currently if you were to pass the empty file with working files the function returns an empty element in the DataFrame (see below). I suggest we change this to only include successful elements

```
'42008': {},
 '46042': {'1996':       YY  MM  DD  hh   WD  WSPD   GST  WVHT    DPD   APD  MWD     BAR   ATMP  WTMP   DEWP   VIS
  0     96   1   1   0  321  10.4  12.7  3.73  16.67  8.30  290  1020.4   14.1  13.6  999.0  99.0
  1     96   1   1   1  326  12.0  15.3  3.70  16.67  8.02  289  1020.2   14.1  13.6  999.0  99.0
  2     96   1   1   2  330  12.6  15.2  3.78  16.67  8.08  293  1020.1   14.1  13.6  999.0  99.0
  3     96   1   1   3  329  12.5  15.8  4.19  16.67  8.48  288  1019.4   14.0  13.6  999.0  99.0
```

To fix this I added an `else` statement to the `try` which will execute on success. Then to specify a nested dictionary e.g. 2 keys we need to use a `defaultdictionary` . I added a test to make sure this works as expected to fix the above behavior.
 I made additional changes to follow PEP8 guidance on max characters per string. See my PR on your fork.

rpauly18:
@ssolson your changes look good to me. If you approve, we can merge this PR.

ssolson:
Yeah is the build is passing then we should merge it. https://travis-ci.org/github/MHKiT-Software/MHKiT-Python

PR #79, opened 2020-11-03
Pecos 0.1.9 

ssolson:
Sets MHKiT requirements to Pecos 0.1.9

Changes window from a kwarg to an arg in check_delta function.


ssolson:
image

PR #91, opened 2021-01-19
Adds PACWAVE resource assessment for EWTEC example

ssolson:
This PR adds the resource assessment work done for the EWTEC paper as an example notebook and adds 2 plotting functions to the `wave.graphics` module. The first is based on the monthly cumulative distribution of energy flux and is detailed in the IEC 62600-101. The second added graphics function is based on the Dunkel et al. paper which bins energy flux is an Hm0, Te, contour and plots the average hours per year each bin occurs for.

image
image


ssolson:
Transferred from https://github.com/SNL-WaterPower/WecOptTool/issues/40
@ryancoe could you comment on if you think this is better as an example or a function?

ryancoe:
Definitely just an example.

rpauly18:
One somewhat issue I have run into is that in the last section, where the representative sea state energy is compared to the total, I do not get the same values as what you cite in the description. 
4     0.9031
8     0.9537
16    0.9866
32    1.0028
64    1.0154
dtype: float64


Is your description incorrect, or is there something else wrong in the analysis?

rpauly18 (approved):

PR #101, opened 2021-03-11
modifying wait times

rpauly18:
Trying to fix the 429 errors

### DIFF
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -821,7 +821,7 @@ class TestWPTOhindcast(unittest.TestCase):
         years = [1990,1992]
         lat_lon = (44.624076,-124.280097) 
         parameters = 'significant_wave_height'
-        #value = randint(300,500)
+        value = randint(300,500)
         #time.sleep(value)
         wave_multiyear, meta = wave.io.hindcast.request_wpto_point_data(data_type,parameters,lat_lon,years)
         assert_frame_equal(self.my_swh,wave_multiyear)
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -821,7 +821,7 @@ class TestWPTOhindcast(unittest.TestCase):
         years = [1990,1992]
         lat_lon = (44.624076,-124.280097) 
         parameters = 'significant_wave_height'
-        value = randint(300,600)
+        value = randint(300,596)
         print(value)
         time.sleep(value)
         wave_multiyear, meta = wave.io.hindcast.request_wpto_point_data(data_type,parameters,lat_lon,years)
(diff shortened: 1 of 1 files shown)


ssolson:
What is it that they changed that made this a problem recently? I mean it had been working for months right?

rpauly18:
I do not think anything changed. We did not have many PRs after we merged the hindcast function. We probably just got lucky during the review process of the hindcast function. Other errors that were thrown during the process likely masked or prevented 429 from ever being thrown. I would say currently ~25% of the time our tests pass without this error being thrown

rpauly18:
My preference at the moment would be to reconfigure the tests so that the hindcast tests are only run with one version of Python. This will take some reworking of the travis.yml file. We could also reconfigure the tests so that the hindcast functions are only run under certain circumstances- but that route is little less straight forward to me.

ssolson:
> I do not think anything changed. We did not have many PRs after we merged the hindcast function. We probably just got lucky during the review process of the hindcast function. Other errors that were thrown during the process likely masked or prevented 429 from ever being thrown. I would say currently ~25% of the time our tests pass without this error being thrown

As far as I can tell every build has failed since you added the new hsds keys commit. I do no think we were lucky before. Am I missing a commit that has passed following the hsds commit?

rpauly18:
Yes, some commits have passed. What I was doing was restarting the build after they passed to see if the changes that were made truly fixed the issue or if it was a case of us just getting lucky. Every time it would fail when I restarted the build.

ssolson:
I do not think we were "lucky" before though. Can we agree that no build ever failed prior to the new hsds commit? It appears to me that there were sufficient commits in this time to see at least 1 build fail.

Can you explain to me what is going on with the commit I just made?

How did I pass Travis but fail coveralls? When I look at the details it is not clear to me how I am failing this.

rpauly18:
I am not really sure. The coveralls check started appearing recently- I do not recall seeing that check in the past. I am guessing because the coverage decreased?

ssolson:
Yep that seems to be the issue. I have modified the coveralls.io settings and I am checking if the build will pass now. 

https://github.com/lemurheavy/coveralls-public/issues/565#issuecomment-120924694

ssolson:
So as a note sleeping between 2 and 10 minutes randomly seems to avoid the issue. I am not going to run the tests using an if statement to only run the hindcast test for Python 3.8.

ssolson:
Interesting. So this worked on the build side but drops our total coverage below 90% because we are no longer testing in 3.7 and 3.9 how frustrating. Okay I will try to do one test in each version,

ssolson:
@rpauly18 I think we are good here. Let's get this merged into the main ASAP.

ssolson (approved):

Issue #102, opened 2021-03-23
Hindcast Multi-param returns non unique index


ssolson:
When calling the WPTO hindcast function for multiple parameters it returns 2 identically indexed DataFrames concatenated together. See the screenshot below for example. So therefore if you requested 2 parameters for 1 day on a 1 hour basis you would end up with a DataFrame of length 48 (not the expected 24).

@rpauly18 is this the intended result or should they share the index?

![image](https://user-images.githubusercontent.com/13438942/112178145-29b9f800-8bbf-11eb-9b5b-67a4f4e4da9b.png)

MWE Code:
```
data_type = '1-hour'
years = [1996]
lat_lon = (44.624076,-124.280097) 
parameters = ['significant_wave_height','energy_period']        
wave_multiparm, meta= wave.io.hindcast.request_wpto_point_data(data_type, parameters,lat_lon,years)

wave_multiparm[~wave_multiparm.energy_period_0.isna()]
wave_multiparm.loc['1996-01-01 00:00:00']
```

ssolson:
fixed in #103

PR #103, opened 2021-03-23
concatenate multi param dfs on same index

ssolson:
I believe this should fix the issue I bring up in #102. So long as all the test still pass (which they did locally) we should be good to go.


ssolson:
It appears that the only 1 version of python per test did not fix the "too many requests" issue. 

Okay so to keep coverage up and test the remote we can run all tests remotely for 1 version of python and add datasets to compare against locally for any other versions we test. I will add those changes here.

ssolson:
Trying to load saved data to compare against doesn't make sense and will not solve the problem as far as I can tell. To just load data would skip running the hindcast function. If we skip running the function then we are don't testing the code and will reduce our coverage anyway.

At this point, I think that the best way to test the code maybe with a minimal call to the function run in all three versions. Remember running the tests in only one python version reduces our reported code coverage when it averages the code coverages across the tested python versions.

Before I try this I think I will add a dedicated wait to the test as they currently stand as this seems to be the best solution that I can currently think of still. Let me know if you guys have other ideas to try or if I am missing something.

ssolson:
lol okay so the wait must be shorter than 10 minutes. One more time.

ssolson:
oh my goodness thwarted by my own tests. Okay so it worked locally bc I run python 3.8 and the multi param is now only tests on python 3.9. This shows a confusing outcome from this hack. I really think the best outcome would be to work with the rex team to set up a wait when it receives a too many call response.

rpauly18:
I believe the error you are getting is because of the change in the returned data from the function if you pass multiple parameters. This should be an easy fix. I can help make the modifications.

PR #104, opened 2021-03-25
Setup reconfig

rpauly18:
This is a first pass at setting up try and except statements around the WPTO hindcast functionality. This first commit is starting small and localized to the import of the hindcast function. If this does not solve the Windows MHKiT-Matlab problem, I will expand this approach to the NREL-Rex dependency itself.


kbrode22:
I tested this pull request in Windows and was able to successfully run MHKiT-MATLAB.

ssolson:
Rebecca, I can run this without issues. I think we should add a warning message of some type instead of just passing on the try-except.

Maybe "WARNING: Wave hindcast was not imported from MHKiT-Python. If you are using windows and calling from MHKiT-MATLAB this is expected."

rpauly18:
Good idea. I will add a warning.

rpauly18:
@ssolson appears there is some kind of issue with Linux and a change on the hsds side that are not playing nice. The HSDS team is debugging, but in the meantime I did get the tests to pass on Linux. I still need to g back and get the other two version of Python working- but how do you feel about making our tests use a mix of Windows and Mac OS? Those are the operating systems we generally support and most of our users will use anyway.

rpauly18:
Still working on this, appears the tests did not run on Windows

rpauly18:
@ssolson I think this is all set now. The h5pyd team needed to make an update to fix a bug that was causing the problem. Take another look at the warning message to see if you think this can be merged.

PR #107, opened 2021-04-22
updating virtual buoy results based on updated data

rpauly18:
The 1 hour data was updated on AWS. Different buoy locations are now in it- so the data changed with the request. The data we are comparing to has been updated accordingly.


rpauly18:
@ssolson nothing changed at all in the code, just the data changed. Hopefully this passes CI

ssolson:
What is weird is that python 3.7, 3.8 passed last time. This passed so yeah I will merge to hopefully remove/ update the fail badge status to pass.

rpauly18:
As an answer to your comment it was only the 1-hour data that changed. Python 3.7 and 3.8 are testing 3-hour data requests.

PR #108, opened 2021-04-22
Feature cdip

ssolson:
The old CDIP PR (#72) was not pulling in the latest commits I made to Kelly's branch so I have started a new one here because I need to see if it is passing the tests. I will update this descriptor more if the old PR continues to not update.


ssolson:
duplicate of #72

PR #110, opened 2021-04-27
WPTO Hindcast directional spectra

rpauly18:
This is a pull request featuring the addition of a function to download the directional wave spectra data from the 1-hour dataset.


rpauly18:
@hivanov-nrel this is the draft PR with a new function for downloading the directional spectra. There are also 2 new wave resource functions for computing spectral JPD's and directional JPD's. We have similar functionality elsewhere in MHKiT for JPD's so you may be able to leverage some of that.

ssolson:
The current JPD is specifically in the Tidal resource module.

hivanov-nrel:
I pushed a recent commit regarding the directional spectrum data. I think for now, it makes sense to have its own function just because of the nature of how that data output is organized. Alternatively, i could squeeze it into the general wpto point data function but would need to add a bunch of if statements to handle the different scenarios of what the user might request. 

I tried to organize the dir wave data into a dict with timestamps and values being a dataframe of freq vs wave dir. It looked nicer and more organized however it take roughly 2.5 hours to reorganize the data this way. I have my solution commented out for now if anyone is curious about the method i was using. If i cant find a more efficient way to do this, i think i would leave the data output in its native state there. 

Also, thanks to Aidan, a function was added to automatically return a predefined region based on the input coordinates fed into it. Aidan provided me with the bounding box coordinates for each region which are hard coded within that function.

mhkit/wave/io/hindcast.py
```diff
@@ -97,4 +135,74 @@ def request_wpto_point_data(data_type, parameter, lat_lon, years, tree=None,
...
+        location metadata for the requested data location   
+    """
+    assert isinstance(lat_lon, (list,tuple)), 'lat_lon must be of type list or tuple'
+    assert isinstance(year,str), 'years must be a string'
+    assert isinstance(tree,(str,type(None))), 'tree must be a sring'
+    assert isinstance(unscale,bool), 'unscale must be bool type'
+    assert isinstance(str_decode,bool), 'str_decode must be bool type'
+    assert isinstance(hsds,bool), 'hsds must be bool type'
+
+    region = region_selection(lat_lon)
```
hivanov-nrel, line 187: Is this the best way to throw this error?

rpauly18: Is this for the region check? You could do an assert that all regions are equal, but you would need to find the region for each passed lat/lon first. This looks like it will only pass back one region.

hivanov-nrel: sorry it looks like that comment was added before my commit got pushed so disregard. Latest commit should have the actual error handling.

rpauly18: The method you are using looks good.

ssolson:
@rpauly18 the more I work with this function the more it looks like we are just creating a wrapper for `np.histogramdd` .

My current code is below. The next step should be to simplify the inputs to just samples and bins...but this is what histogramdd does.  What specifically are we adding that histogramdd does not do?


```
def joint_probability_distribution(x, y, x_edges, y_edges, z=None, 
                                   z_edges=None):
    """
    Calculates the joint probability distribution between two or three
    variables.
    
    Parameters
    ----------
    x: numpy array or pandas Series 
        first variable for JPD
    y: numpy array or pandas Series 
        second variable for JPD
    z: numpy array or pandas Series 
        third variable for JPD        
    x_edges: numpy array  
        first variable bin edges for JPD
    y_edges: numpy array  
        second variable bin edges for JPD
    z_edges: numpy array  
        third variable bin edges for JPD        
        
    Returns
    -------
    jpd: pandas DataFrame
        jpd indexed by x_bins with y_bins columns
    """


    # Calculate the bin centers of the data
    def bin_edges_to_centers(bins):
        centers=[]
        for i in range(bins.shape[0]-1):
            centers = centers.append(mean([var1_bins[i+1], var1_bins[i]]))
        return centers
    x_center = bin_edges_to_centers(x_edges)
    y_center = bin_edges_to_centers(y_edges)
    
    
    if z:
        jpd_data = array([x.flatten(),y.flatten(), z.flatten()]).T
        
        bins=[x_bins,y_bins,z_bins]
        probability, edges = np.histogramdd(jpd_data,bins=bins,density=True)  
        
        z_center = bin_edges_to_centers(z_edges)  
    else:
        jpd_data = array([x.flatten(),y.flatten()]).T
        bins=[x_bins,y_bins]
        probability, edges = np.histogramdd(jpd_data,bins=bins,density=True)
    

    jpd = DataFrame(probability, index=x_center, columns=y_center)

    return jpd
```

hivanov-nrel:
I wrote up a test for the directional spectrum and it looks like the code coverage passed. Let me know if i should write some more to increase the coverage. The only thing left is to output into XARRAY format. I played around with this and its very simple to convert from pandas into XARRAY, however i wrote up the tests to look at the pandas output. Is there similar xarray assert tests that we could use?

rpauly18:
@ssolson you are right, we are not adding much other than making it easy for non-savey Python researchers to calculate JPDs. As I mentioned before, I received several requests for this. It may be enough to have an example in the WPTO hindcast notebook

rpauly18:
@hivanov-nrel there are. A quick Google search came up with a few options, for [example](http://xarray.pydata.org/en/stable/generated/xarray.testing.assert_equal.html).  I would look into what method would work best for this application and what integrates with our tests the best.

ssolson:
> @ssolson you are right, we are not adding much other than making it easy for non-savey Python researchers to calculate JPDs. As I mentioned before, I received several requests for this. It may be enough to have an example in the WPTO hindcast notebook

Okay perfect. I think having an example is the right way to go. I pushed commits Removing the jpd functions from the PR.

hivanov-nrel (approved):

rpauly18:
@hivanov-nrel great job on this. I updated the example notebook. Can you review the changes? I added a plot that is somewhat interactive, can you particularly make sure the description there makes sense and that it works?

hivanov-nrel:
@rpauly18 I ran the example and made a few updates to fix some typos. So for the interactive plot, I was only able to get it to work if I ran it specifically through jupyter notebook. I could not get that plot to work if I ran the notebook inside vscode and I could not get it to work inside jupyter-lab either. Apparently, it does not support interactive matplotlib plots.

PR #113, opened 2021-06-10
Bug series

rpauly18:
Try again with a clean commit history. This is to fix a bug in the recent wave.resource changes that allowed for Series to be passed. There was an oversight that the changes prevented multi-column data frames from being passed- which broke our example notebook.


rpauly18:
@ssolson if you have time, take a quick look at this. I plan to merge this later this morning.

PR #114, opened 2021-06-17
adding peak_to_average

rpauly18:
Peak to average function. 

I had originally started developing this for peak to average power ratio, but it may be better suited as a more general function in utils. This function also still needs tests written


ssolson:
Closing due to inactivity.

PR #124, opened 2021-07-08
adding 3.7 tests back

rpauly18:
Also updating NREL-rex requirements


ssolson:
This line remained the same in the Updated version of rex 

https://github.com/NREL/rex/blob/52db08f1e356f6e704436449dcdd9f105d63b685/rex/multi_year_resource.py#L146

rpauly18:
The change was made in the version that is released on pip. They changed it back on the master branch that was not tagged on pip because they were worried about possible errors associated with the change. The data is stored in linux, so the slashes have to be a certain way

PR #143, opened 2021-11-09
Fix wave elevation function.

### LINKED ISSUE #140: wave elevation function is very slow
```python
elev = surface_elevation(S, time) 
```
is about 20 times slower than 

```python
elev = np.zeros(time.shape)
for ifreq, iS in S.T.iteritems(): 
    elev += iS[0] * np.cos(ifreq*2*np.pi * time + np.random.rand()*2*np.pi)
```

### LINKED ISSUE #142: Time array not used correctly in wave.resource.surface_elevation
Currently the different spectral arrays are being truncated using the start and  end time, e.g. 
```python
phase = phase[start_time:end_time]
```
but these Panda arrays are indexed by frequency (Hz) not time.   It is not doing what it is supposed to be doing.

cmichelenstrofer:
Fixes #140. Fixes #142.

### DIFF
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -62,29 +62,6 @@ class TestResourceSpectrum(unittest.TestCase):
         self.assertLess(errorHm0, 0.01)
         self.assertLess(errorTp0, 0.01)
 
-    # def test_surface_elevation_seed(self):
-    #     S = wave.resource.jonswap_spectrum(self.f,self.Tp,self.Hs)
-
-    #     sig = inspect.signature(wave.resource.surface_elevation)
-    #     seednum = sig.parameters['seed'].default
-
-    #     eta0 = wave.resource.surface_elevation(S, self.t)
-    #     eta1 = wave.resource.surface_elevation(S, self.t, seed=seednum)
-
-    #     assert_frame_equal(eta0, eta1)
-
-    # def test_surface_elevation_phasing(self):
-    #     S = wave.resource.jonswap_spectrum(self.f,self.Tp,self.Hs)
-    #     eta0 = wave.resource.surface_elevation(S, self.t)
-    #     sig = inspect.signature(wave.resource.surface_elevation)
-    #     seednum = sig.parameters['seed'].default
-    #     np.random.seed(seednum)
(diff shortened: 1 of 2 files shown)


cmichelenstrofer:
@ssolson and Rebecca could you review this? 

- Bug fix: 
  - not truncating the spectrum anymore (was being truncated based on time value, but the `pd.Series` was indexed by frequency).
- Speedup 
  - moved calculation of `B` outside loop
   - replaced list comprehension + itertools with `np.outer` function
   - avoided creating a  complex array to then take the real part only, instead create the real part directly.
 - Functionality:
   - changed default seed to `None`. The default behavior should be creating a random phase. A  user can specify a  seed for reproducibility. Should not be the other way around (where we were hardcoding a phase as the default behavior). 
 - Tests:
   - There were two tests that did nothing. They would always pass. Their intention were to check that the default behavior and specifying the seed in the function signature gave the same results. But they always will even if the function is broken.  I commented these out but think they should be deleted. 
 - Examples:
   - I couldn't find an existing notebook that uses the `surface_elevation` function. If there isn't one, maybe we should add it to an existing example. 
 - Reference
   - We need to find a good reference for this function (converting frequency domain to time domain)

rpauly18:
@cmichelenstrofer happy to review this. Are you still working on the fix for why the tests are failing, or would you like for me to look into this?

cmichelenstrofer:
@rpauly18 I just fixed the tests. You can go ahead and review it. Thanks!

rpauly18 (approved):
Looks good. Are you still working on more changes, or is this ready to be converted from a draft?

cmichelenstrofer:
@rpauly18 just changed it from '_draft_' to '_ready to review_'.  I deleted the two obsolete tests that I had commented out.

cmichelenstrofer:
Something wrong with the tests. They were passing before and nothing changed other than deleting comments. Seems the issue is not with this PR but with accessing some data at NREL needed for the tests.

rpauly18:
@cmichelenstrofer random failures happen sometimes because of the WPTO hindcast server being overloaded with request. I restarted the tests. If they fail again, I will investigate further.

cmichelenstrofer:
@rpauly18 can this be merged?

PR #152, opened 2022-02-11
NDBC Metocean data

akeeste:
This PR adds the ability to read continuous wind data to the NDBC I/O module. It does the following:
- [x] adds the "cwind" variable to each NDBC function documentation
- [x] jupyter notebook example (`wind_example`) to read and visualize wind data
- [x] adds test to read continuous wind data files, both with and without units
- [x] adds a check in ``waves.io.ndbc.request_data()`` to catch and remove units from numerical data.


examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "# Get average hourly wind speeds\n",
+    "hours = np.arange(1,24,1)\n",
+    "hourly_mean_speeds = []\n",
+    "for hour in hours:\n",
+    "    # Create a boolean mask using the start and end times\n",
+    "    start_time = '2018-01-11 ' + str(hour-1) + ':30:00'\n",
+    "    end_time = '2018-01-11 ' + str(hour) + ':30:00'\n",
+    "    mask = (daily_data.index >= start_time) * (daily_data.index < end_time)\n",
+    "    hourly_data = daily_data[mask]\n",
+    "    hourly_mean_speeds.append(np.mean(hourly_data.WSPD))"
```
ssolson: I believe this whole code block is simply :
`daily_data.resample('h').mean()`

examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "fig = plt.figure()\n",
+    "ax = fig.add_subplot(111)\n",
+    "ax.set_xlabel('Time (h)')\n",
+    "ax.set_ylabel('Speed (m/s)')\n",
+    "ax.set_title('Hourly mean wind speeds on January 11, 2018')\n",
+    "line = ax.plot(hours,hourly_mean_speeds,'o')"
```
ssolson: what do you think about adding gridlines to the plot?

examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
+     },
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "monthly_data = ndbc_data['2018']['2018-01']\n",
+    "daily_data = monthly_data['2018-01-10']\n",
```
ssolson: Because you need to repeat the month ('01') in the monthly data and daily data I would suggest making this yearly data then daily data or simply:
`daily_data = ndbc_data['2018']['2018-01-10']`

examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "\n",
+    "# Remove bad speed data\n",
+    "mean_speed_data = daily_data[daily_data.WDIR < 999]\n",
```
ssolson: This operation is okay. I think this would be better:
`daily_data['WDIR'] = daily_data['WDIR'].replace({999.0:np.NaN})`

I think you should mention the different NaN values when they are first shown up above on the initial import. Maybe replace them at that time.

Also, you change the name to a mean but do not take a mean, is this correct?

examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
...i7YEtj94MT9f+7Xpa/eU3KG9qBgDcNW0ialrUWDs7FZE+tl/bdDjzPIwmCwghyDtff9kEBgDajKaOOloUBvPQLig4ENXV1SgtLUVycvJF/bGGAmslMbm5ubjpppvqlErlHEppqxVDZP6IDcCwbdQDRSltUyqVs++6666qEydO9PnxQzl56eTi4oJp06bBYrEgIyMDer3e3iE5tNHBPhAQAqmTGAKBADylIATQGdvLY3y0NwMWjodcpcVv+eX9vk6dSg2jhQOlwOfHs/BjXiHWbv++5wf+gc5kxoZ96Xj/f8dhsvQuwbhz6VQICIGflyumjLv8VmxCCDbdvQSpMeF4YvFsjPKzfXJVVtOIJ97aiU92HL1ou7mtmM1mZGdnQ6FQYPr06UMueek00C3WFRUVWLhwoVwul8+nlNbZIETmAoRSq98cDRuBsQJKaRMhZM7SpUszDh48GBQX17tKoY6QvHQSCAQYO3YslEolTpw4gcjISISHh7PRmH6YEh+Oz9avRKNKBy8PKT796QQmjwlFUlR7c8TZiVH49lguBIRg3Kj+D+e/dcNCvPXrb4jy9cLh0goQQiDuYzNAAPjnL+nYnnm2qx7LmrlTL3vsvvJiPHboJwS7eWDXZ/fDz8Wtx/OnjglH6pjwPsfVX8+9+yPKa5qQVVCN+NGBmDnROut+uqNQKJCfn4+YmBiEhIQM+d+X/o7ENDY2Yv78+Y21tbXXU0qLbRkjw3RiCYyVUEqrCCF/+tOf/nTg+PHjfiEhV16M6EjJy4X8/Pwwffp0FBQUoL6+HklJSZBKh0Zht8HA8Txe+ewXZJ+rxsM3z8KCqf2rCTN21O9bgt9du/Si+55cOhtLpybAx90VPu79Xx8+OyYSs2Pat+buzilAXr0Cd6Wl9Pk8/AWfzLgeRiz+cSodOosZNRoVfq0sxar4pD5fDwA+LzmCirZGPBG/EB5O1n19ublIIBQQUErh5mKbvlFmsxl5eXkwmUxIS0vrtkHnUNXXJEar1WLu3LnNtbW1t1NKTw5GjAzYIl6wBMaqKKV5IpHohrlz5+46ceKEz+XqqDhq8tJJLBYjKSkJSqUSGRkZGD16NMLCwob8p8v+snA83vzkV5wrbcD1C8bh4Mnz0BvN+PsX+/udwFwJIQRjgq3bf2hJ0lgsSRoLAMhTytGs12NGWES37QUUei3azCZEerRv7X3
```
ssolson: Again I would recommend using the replace function here

ssolson: I don't really think `daily_data` is a good name here. Shouldn't it be `Jan_data`?

ssolson (commented):
Hey Adam great job on this. I have just a couple minor comments to kick off the review.

Overall I think you explain the use of the NDBC model much better than I did in the Pacwave example. Therefore I think I will remove my description of using wave.io.ndbc and point to this example.

@rpauly18 this example is the basis of the metocean module. It is currently called `wind_example`. I think we could leave this as wind example for now or rename to `metocean`. If we don't like either of those for now this could be renamed NDBC example as we do not have one specifically for that wave.io submodule. What is your thoughts/ preferences?

rpauly18:
@ssolson I prefer "metocean" over "wind"

examples/wind_example.ipynb
akeeste: Good call, I'll add a section about replacing the NaN values with np.NaN and do it up front before creating separate daily and monthly datasets.

akeeste:
@ssolson @rpauly18 
I made the requested changes to this PR including:
- [x] Move graphics import from the example to the ndbc module
- [x] Fix capitalization and other minor grammar
- [x] Add grid lines to time series plot
- [x] Renamed the variables to Jan_data and Jan_11_data for clarity, and replaced the hour sampling with a call to resample() for simplicity
- [x] Replaced 99, 999, 9999 with NaN up front and clarified the need to do so

examples/wind_example.ipynb
akeeste: Agreed, I renamed the daily/monthly variables to ``Jan_data`` and ``Jan_11_data`` for clarity

examples/wind_example.ipynb
akeeste: Good call, it makes the plot easier to read

examples/wind_example.ipynb
akeeste: Thanks! I wasn't aware of this dataFrame method yet. Updated this to use ``resample()``

examples/wind_example.ipynb
```diff
@@ -0,0 +1,925 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from mhkit.wave.io import ndbc\n",
+    "from mhkit.tidal.graphics import plot_rose \n",
+    "from mhkit.tidal.graphics import plot_joint_probability_distribution\n",
```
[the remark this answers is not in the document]
akeeste: I moved these imports out of the example and into the ndbc module, this seemed most similar to how the tidal module inherited from the river module

examples/wind_example.ipynb
akeeste: Along with renaming these variables for clarity, I made Jan_data and Jan_11_data reference the yearly dataFrame directly

PR #163, opened 2022-03-25
regenerate WDRT examples

hivanov-nrel:
Sorry for the delay......this PR is to re-run the WDRT example notebooks so the full output is displayed.


cmichelenstrofer:
@hivanov-nrel Thanks for doing this! I really need to figure out a way to pull buoy data on my work computer... I spent some time on it but couldn't figure it out. Need to revisit it.

PR #166, opened 2022-04-04
request_wpto_directional_spectrum improvements

ryancoe:
Update the `wave.io.hindcast.request_wpto_directional_spectrum` function to:

1. Use `datetime` objects
2. Create a `DataArray` (not `Dataset` with weird variable naming)

### DIFF
--- a/mhkit/tests/test_wave.py
+++ b/mhkit/tests/test_wave.py
@@ -1045,7 +1045,7 @@ class TestWPTOhindcast(unittest.TestCase):
             meta_dir)=(wave.io.hindcast
                        .request_wpto_directional_spectrum(lat_lon,year='1995'))
             dir_multiyear = dir_multiyear.sel(time_index=slice(dir_multiyear.time_index[0],dir_multiyear.time_index[99]))
-            dir_multiyear = dir_multiyear.rename_vars({87:'87',58:'58'})
+            dir_multiyear = dir_multiyear.rename({87:'87',58:'58'})
 
             assert_frame_equal(self.ml,wave_multiloc)
             assert_frame_equal(self.ml_meta,meta)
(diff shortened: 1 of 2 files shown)


ssolson:
Hey @ryancoe I am less familiar with this module as I am not the author. The author Rebecca is expected to be on an extended leave anytime now. Your change to a Dataarray broke the way that multiyear arrays were working. Could you take a quick look and let me know if you would set up the multiyear storage differently based on your proposed change here?

```
======================================================================
ERROR: test_multi_loc (mhkit.tests.test_wave.TestWPTOhindcast)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/MHKiT-Python/MHKiT-Python/mhkit/tests/test_wave.py", line 1075, in test_multi_loc
    dir_multiyear = dir_multiyear.rename_vars({87:'87',58:'58'})
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/xarray/core/common.py", line 239, in __getattr__
    raise AttributeError(
AttributeError: 'DataArray' object has no attribute 'rename_vars'
```

ryancoe:
@ssolson - I think `test_wave.py` has some odd subtleties that I didn't understand (and that caused the tests to run very slowly but pass on my machine).

1. It runs different tests for different version of Python (I think this is to avoid making too many requests to the HSDS server that `rex` uses, which has a daily limit
2. It uses long `time.sleep` statements to avoid making requests to HSDS at the same time.

Can you confirm or deny this interpretation ;) ?

Another question: Is there a reason the data variables have names like `87` and `58`? I took this to be an oversight (i.e., `87` should be `S` for _spectral density_), but maybe I was missing something?

ssolson:
Closing in favor of #220

PR #169, opened 2022-05-04
DOLfYN source code

jmcvey3:
Integrated source code with latest bugfixes.
Will need to update jupyter notebook documentation links once docs are completed.


ssolson (commented):
James thank you for putting this together. I feel like I have a quarter of the knowledge and time required to understand and test the many capabilities that DOLfYN brings to MHKiT. I have done my best to provide feedback and questions where it felt appropriate. I have some questions around the documentation and what some the formatting in the docstring looks like in our API documentation outstanding. Let us have a chat soon and get this PR wrapped up. Thanks again!

jmcvey3:
@ssolson Did some renaming and reorganization to hopefully make the analysis functions easier to understand. Top level functions are located in velocity.py, mid-level in binned.py, and base-level in the "tools" folder. I would expect most users to utilize the top-level functions, but the public functions in the lower two levels are useful for python-knowledgeable users.

We had also discussed creating an issue to reassign functions that exist in other packages (like scipy) as I was under the impression that generic statistical calculations like coherence, phase-angle and auto/cross-covariance would exist in other packages. Apparently they do not.

jmcvey3:
Failures appears to be new bugs in the netCDF4 1.6.0 release

ssolson:
Hey James I have been looking into this more and it appears that it is actually the Dolfyn. xarray `.to_netcdf ` method which is breaking the UNIX tests for MHKiT which you can checkout on #181. I am going to be poking around to see if I can fix it but if you get a chance to take a quick stab at it I would appreciate your thoughts on the problem and could run with any ideas you may have to fix it.

Based on our last conversation/ review of Dolfyn is this ready to go?

jmcvey3:
> Hey James I have been looking into this more and it appears that it is actually the Dolfyn xarray `.to_netcdf ` method which is breaking the UNIX tests for MHKiT which you can checkout on #181. I am going to be poking around to see if I can fix it but if you get a chance to take a quick stab at it I would appreciate your thoughts on the problem and could run with any ideas you may have to fix it.
> 
> Based on our last conversation/ review of Dolfyn is this PR ready to go?

It is ready to go. I haven't been able to replicate this error yet, but it's strange because it hasn't been an issue before

jmcvey3:
> > Hey James I have been looking into this more and it appears that it is actually the Dolfyn xarray `.to_netcdf ` method which is breaking the UNIX tests for MHKiT which you can checkout on #181. I am going to be poking around to see if I can fix it but if you get a chance to take a quick stab at it I would appreciate your thoughts on the problem and could run with any ideas you may have to fix it.
> > Based on our last conversation/ review of Dolfyn is this PR ready to go?
> 
> It is ready to go. I haven't been able to replicate this error yet, but I'll keep searching for a solution

@ssolson Apparently recreating the test file fixes that issue #181

PR #185, opened 2022-08-15
Updates for WEC-Sim v5.0

akeeste:
This PR updates the wave/io/wecsim module to include updates related to the recent WEC-Sim v5.0 release:
- updates WEC-Sim .mat files with data from v5.0
- adds the ability to read the WEC-Sim cable class


ssolson:
Adam thank you for being proactive on this! 

- I think the answer is yes but to be sure is this still backward compatible with prev5.0 WecSim files?
- It looks like the version of the WecSim example notebook you are pushing has not been run. Could you run the notebook locally and then make a commit so the results are shown?
  - Secondly on the notebook, is there anything we could do to use the data you uploaded?
- Lastly, it looks like you added new data but did not add a test for the data. Do you think you could add a check for the cables into the tests?

akeeste:
@ssolson In response to your questions above:
* WEC-Sim v5.0 is not backwards compatible, but most of those changes are for input parameters, not the outputs that MHKiT reads. There is one output name that changed and will not be backwards compatible. Since this is a small change, I can add a try-catch, so that MHKiT's WEC-Sim module remains backwards compatible
* Good catch, the notebook is updated now. 
* I left this in largely the same format as before. Of the three datasets, one is used in the notebook example. All three are used for the tests. 
* I will add/swap one dataset with a cable example so the tests can check it

My TODO
- [ ] add try-catch
- [ ] update notebook with results
- [ ] add cable dataset to tests
- [ ] add checks for the cable class to tests

ssolson:
@akeeste I reviewed each of your changes and it looks like you got everything. Can I merge this?

akeeste:
@ssolson Yes it's ready. Let's merge

ssolson (approved):

Issue #167, opened 2022-04-06
Metocean Module


ssolson:
@rpauly18 do you have suggestions for data or functions beyond pulling in NDBC wind data for our metocean module?

@akeeste and I were discussing what direction to head after PR #152 is merged.

akeeste:
Noting that we also had some discussion on using [Argo buoys](https://argo.ucsd.edu/) to get temperature and salinity over the entire water column. However these buoys drift over time which might pose some difficulties.

PR #186, opened 2022-08-16
Dolfyn general updates

jmcvey3:
Fixes for:
- ADV skipped timesteps
- ADV attributes
- proper setting of max gap in ADV cleaning function
- error in integral length scale calculation
- error in despiking function
- move Reynolds stress and cross spectral density calculations to ADV folder
- prevent inaccurate TKE calculation from ADCP velocity input

I was updating this branch from master, so there may be repeated commits in this list, though that shouldn't matter after resolving conflicts and merging.

PR #181, opened 2022-07-25
NetCDF4

ssolson:
This PR enables MHKiT to work with the latest version of NetCDF4 (#177).

To fully pass the test on python 3.7, 3.8, 3.9 we use a conda environment (windows and py 3.7 were the only version combinations that could not find the hdf5 path). I have added separate test for pip versions 3.8, 3.9. This expands our tests to check for behavior when users use both pip and conda. This addition is in line with the testing consultation on MHKiT summarized in #183. 

Separately on netcdf 1.6.0 there is a new compression kwarg. For Dolfyn I have removed the compression key on the io test. This test will be fixed in #186.

Issue #61, opened 2020-08-18
Ordering lake shoreline data


scjames:
When generating a flexible-mesh model, most grid-generation software (e.g., RGFGRID) require that lake shoreline data be ordered so that the boundary is swept continuously in the clockwise direction (even if this yields segments with convex curves). A quick Google search was not productive; it was difficult to find a tool/utility for what would seem a common request. MATPLOT has a "boundary" function that might work (but a MATLAB license is required) and there is Alpha Shapes that might also work (the source code was not readily available). I would like to see you develop a utility code (especially if it were in Python) that facilitates conversion of a raw, arbitrarily sorted, two-column ASCII (or *.shp) file into ordered, clockwise, shoreline data. An important feature to include would be the ability to handle islands within the lake. 

I have attached an example dataset that includes convex boundary segments and islands, so this would be a challenging example problem and if you can get this working, it should work for most other datasets, too.

[WatereeShoreline.zip](https://github.com/MHKiT-Software/MHKiT/files/5091954/WatereeShoreline.zip)

ssolson:
@scjames thanks for your interest in MHKiT and submitting an issue. As discussed offline I and others at Integral have run into this issue when creating land boundary files for setting up simulations and created a python script that has done some of this that we could adapt to solve this common problem. I believe MHKit is a good fit for this type of function and we have some work related to this coming in Q1. I will assign this task to myself to get implemented in Q1 as soon as I knock out some of the previous open issues!

ssolson:
Below is an arbitrary alpha shape function and plotting script. As can be seen, shoreline detail is currently being cut off. Increasing the detail leads to many smaller individual polygons. The concavity varies significantly throughout the shoreline and makes a single alpha parameter non-ideal. A couple of ideas I would want to try:

1. partitioning the data into smaller sections and considering individual alphas for each
2. considering a mean or expected difference between points and trying to only increase this distance as needed to get to the next point. In this case, most of the data seems consistently spaced. The southeast area is an example problem area of non evenly spaced data. Also, there are  areas where bridges go over the lake and could cut off the shoreline data following this approach

```
import matplotlib.pyplot as plt
import pandas as pd
import numpy as np
from descartes import PolygonPatch
from shapely.geometry import Point, MultiLineString, MultiPoint
from scipy.spatial import Delaunay
from shapely.ops import  polygonize, cascaded_union

def add_edge(edges, i, j):
    """
    Add a line between the i-th and j-th points,
[...]

PR #187, opened 2022-08-16
Metocean module - WIND Toolkit

akeeste:
@rpauly18 @ssolson 
This PR builds on a metocean module discussed in #167 by adding the IO capability for WIND Toolkit data, including vertical profiles of:
* wind speed
* wind direction
* temperature
* pressure

as well as surface values of 

* inverse Monin-Obukhov Length
* Precipitation rate
* Relative humidity

@Alex-McVey and I have not heard from Senu about desired metocean variables yet. However I have been working to build out IO functionality to pull in offshore WIND Toolkit data. This is going well so far as it is similar to the WPTO hindcast module. A new metocean example and tests for this module are in progress.

TODO items:
- [X] Build functionality to get WIND Toolkit data from AWS
- [X] Update metocean example to use and compare both NDBC and WIND data
- [X] Create tests for WIND Toolkit functions
- [x] Create short function to easily format the parameters for speed, direction, temperature profiles to decrease user burden


mhkit/tests/wave/io/test_wind_toolkit.py
```diff
@@ -0,0 +1,183 @@
...
+            dtype = {'pressure_200m_0':'float32'})
+        self.my.index = pd.to_datetime(self.my.index)
+
+        self.ml = pd.read_csv(join(datadir,'wtk_multiloc.csv'),
+            index_col = 'time_index',
+            names = ['time_index','windspeed_10m_0','windspeed_10m_1'],
+            header = 0,
+            dtype = {'windspeed_10m_0':'float32',
+                     'windspeed_10m_1':'float32'})
+        self.ml.index = pd.to_datetime(self.ml.index)
```
ssolson, line 53: Adam here you explicitly state header names, index, and data types. It seems to me that the csv files have all this information and pandas will interpret almost all of it. What is your thought process around the specificity of the imports? Do you think it will prevent future issues? Do you think the specificity of the imports could cause us to make incorrect assumptions if the format changes in the future or if users are less specific in their imports?

akeeste: Hi @ssolson 

Overall I'm not tied strongly to these import parameters. The `pd.read_csv` calls works very similarly without them. I can remove them if you feel strongly. See below for specifics:

> What is your thought process around the specificity of the imports? 

I largely based this test's formatting off of [mhkit/tests/wave/io/test_hindcast.py](https://github.com/MHKiT-Software/MHKiT-Python/blob/Develop/mhkit/tests/wave/io/test_hindcast.py), which also specifies these input. If there is not a specific reason that test was done that way, I am fine changing this one. 

>Do you think it will prevent future issues? Do you think the specificity of the imports could cause us to make incorrect assumptions if the format changes in the future or if users are less specific in their imports?

These import parameters are defined so that `pd.read_csv` gives the same metadata that the MHKiT function does by default. Since this `read_csv` call is only used in the tests, it will not affect users' work. The dataset format appears to be static, but if it did change then it could break this test. On the other hand the test breaking would alert us that something in the WIND Toolkit formatting has changed and we might need to check our functionality

ssolson: Okay I see. Yeah let us keep it as is, I simply wanted to understand if there was an explicit reason for laying out the tests this way that I was overlooking. Thanks!

ssolson:
Adam today I was looking at the wtk inputs and I think we should have a function to format the inputs for the user.

E.g. the inputs currently;
```
wtk_inputs = {'time_interval':'1-hour',
              'wind_parameters':['windspeed_10m','winddirection_10m'],
              'temp_parameters':['temperature_2m','temperature_20m',
                                 'temperature_40m','temperature_60m',
                                 'temperature_80m','temperature_100m',
                                 'temperature_120m','temperature_140m',
                                 'temperature_160m'],
              'year':[2018],
              'lat_lon':(40.748, -124.527)}
```
I think a quick function that returned the formatted list for wind direction and temperature should be included in the WIND toolkit. Something like:

```
wind_parameters = wind_toolkit.wind_parameters(windspeed=[10], winddirection=[10])
temp_parameters = wind_toolkit.temp_parameters([2,20,40,60,80,100,120,140,160])

wtk_inputs = {'time_interval':'1-hour',
              'wind_parameters': wind_parameters ,
              'temp_parameters': temp_parameters ,
              'year':[2018],
              'lat_lon':(40.748, -124.527)}
```

Alternatively, we could just have the inputs converted on the backend allowing the user to input more human-readable/ intuitive inputs like:
```
wtk_inputs = {'time_interval':'1-hour',
              'wind_parameters':[(10,10)],
              'temp_parameters':[2, 20, 40, 60, 80, 100, 120, 140, 160],
              'year':[2018],
              'lat_lon':(40.748, -124.527)}
```
Where the wind parameter speed and direction have been treated as a tuple and the temp parameters as a list. I don't expect either of the above to be the perfect solution but they are suggestions I was thinking of to make using the tool easier. What do you think?

akeeste:
@ssolson Agreed, they can be tedious to type out and we can do something to make it easier. I prefer adding another small function to create variable for speed/direction/temperature profiles. That should require less checks for the parameters `wtk_request_point_data` and keep it a bit cleaner.

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+    Returns
+    -------
+    region : string
+        Name of predefined region for given coordinates
+    '''
+    assert isinstance(lat_lon, (list,tuple)), 'lat_lon must be of type list or tuple'
+    assert isinstance(lat_lon[0], (float,int)), 'lat_lon values must be of type float or int'
+    assert isinstance(lat_lon[1], (float,int)), 'lat_lon values must be of type float or int'
+    assert isinstance(preferred_region, str), 'preferred_region must be of type string'
+
```
ssolson, line 28: `assert len(lat_lon)==2`

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+        elif preferred_region == 'NW_Pacific':
+            region[0] = 'NW_Pacific'
+        else:
+            raise TypeError(f"Preferred_region ({preferred_region}) must be 'Offshore_CA' or 'NW_Pacific' when lat_lon {lat_lon} falls in the overlap region")
+        
+    if len(region)==0:
+        raise TypeError(f'Coordinates {lat_lon} out of bounds. Must be within {rDict}')
+    else:
+        return region[0]
+
```
ssolson, line 53: 2 lines between all functions

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+def region_selection(lat_lon, preferred_region=''):
+    '''
+    Returns the name of the predefined region in which the given coordinates reside.
+    Can be used to check if the passed lat/lon pair is within the WIND Toolkit hindcast dataset. 
+
+    Parameters
+    ----------
+    lat_lon : list or tuple
+        Latitude and longitude coordinates as floats or integers
```
ssolson, line 14: must be of length 2 correct? 

Seeing a type `list` it is not immediately clear that I cannot pass a list of tuples and this function would return multiple regions.

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+            region[0] = 'NW_Pacific'
+        else:
+            raise TypeError(f"Preferred_region ({preferred_region}) must be 'Offshore_CA' or 'NW_Pacific' when lat_lon {lat_lon} falls in the overlap region")
+        
+    if len(region)==0:
+        raise TypeError(f'Coordinates {lat_lon} out of bounds. Must be within {rDict}')
+    else:
+        return region[0]
+
+def plot_region(region,ax=None):
```
ssolson, line 54: I think we should add this to the example notebook. Seems like a cool function to highlight.

ssolson: What do you think about passing the lat_lon value instead of the region?  I think showing a lat_lon value in the plot would be useful for helping the user understand where their lat lon is within the region.

I also think plotting multiple regions could be useful, especially for the overlap region. 

image

I do not feel strongly about expanding this function, I just like plotting functions.

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+    
+    Returns
+    ---------
+        ax : matplotlib pyplot axes
+    '''
+    assert isinstance(region, str), 'region must be of type string'
+    assert region in ['Offshore_CA','Hawaii','Mid_Atlantic','NW_Pacific'], f'{region} not in list of supported regions'
+    
+    wind_path = f'/nrel/wtk/'+region.lower()+'/'+region+'_*.h5'
+    # wind_path = f'/nrel/wtk/'+region.lower()+'-5min/'+region+'_*.h5'
```
ssolson, line 75: No longer needed?

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+    if ax is None:
+        fig, ax = plt.subplots()
+    ax.plot(lons,lats,'o')
+    ax.set_xlabel('Longitude (deg)')
+    ax.set_ylabel('Latitude (deg)')
+    ax.grid()
+    ax.set_title(f'Extent of the WIND Toolkit {region} region')
+    
+    return ax
+
```
ssolson, line 94: 2 lines between functions

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+    assert parameter in ['windspeed','winddirection','temperature','pressure']
+    
+    parameter_list = []
+    for e in elevations:
+        parameter_list.append(parameter+'_'+str(e)+'m')    
+    
+    return parameter_list
+
+
+
```
ssolson, line 130: 2 lines between functions

ssolson (commented):
Hey, Adam great work on this. I just have a couple of ideas, comments, and questions.

mhkit/wave/io/wind_toolkit.py
```diff
@@ -0,0 +1,251 @@
...
+            Data set type of interest
+            Options: '1-hour' '5-minute'
+        parameter: string or list of strings
+            Dataset parameter to be downloaded
+            5-minute dataset options: 
+                'windspeed_10m', 'windspeed_20m', 'windspeed_40m', 'windspeed_60m', 'windspeed_80m',
+                'windspeed_100m', 'windspeed_120m', 'windspeed_140m', 'windspeed_160m', 'windspeed_180m', 'windspeed_200m', 
+                'winddirection_10m', 'winddirection_20m', 'winddirection_40m', 'winddirection_60m', 'winddirection_80m',
+                'winddirection_100m', 'winddirection_120m', 'winddirection_140m', 'winddirection_160m', 'winddirection_180m', 'winddirection_200m',
+            1-hour dataset options: 'time_index', 'coordinates', 'inversemoninobukhovlength_2m',
```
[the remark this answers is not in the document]
akeeste: I expected this to just return the ``time_index`` variable, kind of an empty call without any more data. I agree it's not particularly useful, but the resource extraction tool claims to allow it so I kept it for completeness.

The WIND Toolkit data is accessible from a few different locations (AWS, NREL API) and contains several different datasets. The offshore ones are allowed here, but the original WIND Toolkit dataset is onshore data from the continental US. It is a bit unclear to me how the offshore data may differ from the original onshore dataset. Their documentation is not particularly clear about possible differences. I'll double check the variables that are/are not allowed and confirm that they're the same for the different offshore regions.

mhkit/wave/io/wind_toolkit.py
akeeste: Good point. I prefer to keep it as a single lat_lon pair, to avoid multiple sizes in ``preferred_region``. I'll remove list as an option so that this is clear

mhkit/wave/io/wind_toolkit.py
akeeste: Sure I can add this to the example notebook, it's a good idea to make sure users know it exists. 

Do you know a way to isolate the boundary points from a list? The variable is just downloaded as a massive list of all points in the region, sometimes in the range of 100,000-1,000,000 points. It would make it a lot easier to plot multiple regions and overlap with just the boundaries.

ssolson: Yes 😄 actually in #61 I posted some code that does just that although it was never implemented it would work as needed for this case. I was using an alpha shape approach. If you are interested in other approaches the top answer from [this](https://gis.stackexchange.com/questions/5426/finding-boundary-co-ordinates-from-given-set-of-point-co-ordinates) question has what appears to be a good summary of other options. 

Its always best to see if these exist in any of our current MHKiT dependencies but alpha shape did not seem to be there from quick look.

akeeste: @ssolson I updated the notebook to highlight each of the provided functions, including the plotting one. In ``plot_region()`` users can now specify a region and their specific lat-lon point to show on top of that region. I rethought the boundary idea, and implementing robust alpha shapes seems like a lot of work and potential upkeep. I left the plots as is, so they look like a solid-colored block of that region, instead of the outline. I guess that it would always make the plotting function more expense, but have not checked. But because I left that out, plotting the overlap is hard. Each region would just lie on top of the other without clear boundaries. Users could instead use ``region_selection`` to check for the overlap region, or use ``plot_region`` twice to show both the California and the NW Pacific regions. 
image

akeeste:
Hi @ssolson

I have finished the various clean-up items and updated the notebook to reference each WIND Toolkit helper function. The only outstanding item is getting your thoughts on the updated plotting function, see your [review comment](https://github.com/MHKiT-Software/MHKiT-Python/pull/187#discussion_r980471352) that is not resolved. 

Let me know if there's anything else we could improve, otherwise I'm pretty happy with things.

mhkit/wave/io/wind_toolkit.py
ssolson: I think highlighting the point is pretty cool. Emily and I will be putting alpha shapes in for D3D plotting so if there is any interest in revisiting the outline at a later date we can.

mhkit/wave/io/wind_toolkit.py
ssolson: What did you think about adding this assert?

mhkit/wave/io/wind_toolkit.py
ssolson: I still think this needs a second line.

mhkit/wave/io/wind_toolkit.py
ssolson: Do we need the commented out wind_path?

ssolson:
Adam apologies. I thought the comments I had made were showing up on your most recent commit for me to review the change but they are on the old commit.  I will resolve those comments again be cause I see the issue is resolved on your most recent commit.

akeeste:
@ssolson Yes I should have pushed all those small changes. Thanks for checking. 

A couple tests related to the wind_toolkit functions aren't passing. I'll double check this locally and see what's going on.

akeeste:
This PR has some issues with the tests. The fact that different tests are failing in different jobs indicates to me that this may be an issue with the number of requests we are sending to the NREL developer server on AWS. 

My tests pass locally, but on GitHub I see the following failures:
- Conda-Ubuntu-3.9 - ``plot_region()`` test fails
- Conda-MacOS-3.8 - ? I can't view details in Actions
- pip-MacOS-3.8 - WPTO Hindcast tests fail (also pulling from AWS)

These are in addition to pip-windows-3.8/3.9, but I believe those are a different problem. @ssolson thoughts?

ssolson:
@akeeste I believe the additional requests made to the AWS server are causing some of the issues. I re-ran the tests and all of the rex failures went away which confirms this. However, this PR is having issues with pip install on Windows 3.8/3.9 which other PRs are not currently having issues with e.g. #190 is currently passing without this issue. I cannot see any additional HDF5 or h5py changes in this PR which would cause these issues. This seems related to #181 but again I am not sure what has changed to cause pip install to fail. Adam are there any new packages or uses related to NetCDF/ hdf5/ h5py that I am currently missing in this PR?

ssolson:
@akeeste I see that #190 was pointed against the master branch which did not include #181 which caused those pip changes. I am going to work on a fix today for this PR and #186 which if currently failing for the same reason.

ssolson:
Apologies if my follow-up was not clear. I have confirmed now that the issue is the pip install/ the new version of NETCDF on the GitHub actions server. I am working on a fix in #194.

akeeste:
Since #211 has fixed our HSDS issues, I'm redoing these tests. I'll confirm that the new HSDS calls for metocean data are not causing issues, then we can merge this PR.

ssolson:
Looks like and easy fix for the failing tests
`FAILED ::TestWPTOhindcast::test_multi_loc - NameError: name 'np' is not defined`

akeeste:
> Looks like an easy fix for the failing tests `FAILED ::TestWPTOhindcast::test_multi_loc - NameError: name 'np' is not defined`

Agreed. I'm  not sure when/why I took that line out, but it's back in. 

A couple last things--reviewing the test structure one more time I see that:
1. WIND Toolkit tests are still in the main Actions workflow, not with the hindcast workflow or it's own workflow
2. WIND Toolkit tests still use the previous hindcast test layout where certain function tests are split up by Python version (i.e. python 3.8 only uses ``test_multi_loc``, 3.7 only uses ``test_multi_year``)

The WIND Toolkit functions use the same rex and HSDS dependencies as the WPTO hindcasts. Breaking out a WIND Toolkit workflow might alleviate calling the server too much, but calling the tests for every python version will do the opposite. @ssolson after working on #211, do you have an intuition about what's best for us right now?

ssolson:
> 

@akeeste  please see my PR at https://github.com/akeeste/MHKiT-Python/pull/1

ssolson:
@akeeste I specified the hindcast tests from the typical tests as if they were in the new test module causing a coverage issue. I have fixed that mistake by excluding the correct files in my most recent PR here https://github.com/akeeste/MHKiT-Python/pull/2

akeeste:
@ssolson I merged your PR https://github.com/akeeste/MHKiT-Python/pull/2 and also updated ``coveragehindcast.rc`` to refer to the correct files.

ssolson:
@akeeste I believe merging akeeste#3 should get the tests passing.

akeeste:
@ssolson tests are passing! I will merge while there's no conflicts and issues.

PR #193, opened 2022-09-22
Add function to compute value at a given return period

mbruggs:
This PR adds a function called `return_year_value`. An example usage can be seen in the `extreme_response_full_sea_state_example.ipynb`. The example would be adjusted from:

image

To:
image

With the `return_year_value` function defined as follows:
```
def return_year_value(ppf, return_year, short_term_period_hr):
    """
        Calculate the value from a given distribution corresponding to
        a particular return year.
        Parameters
        ----------
        ppf: callable function of 1 argument
            Percentage Point Function (inverse CDF) of short term distribution.
        return_year: int
            Return period in years.
        short_term_period_hr: int, float
            Short term period the distribution is created from in hours.
        Returns
        -------
        value: float
            The value corresponding to the return period from the distribution.
    """
    assert callable(ppf)
    assert isinstance(return_year, int)
    assert isinstance(short_term_period_hr, (float, int))

    p = 1 / (return_year * 365.25 * 24 / short_term_period_hr)

    return ppf(1 - p)
```

### DIFF
--- a/examples/extreme_response_full_sea_state_example.ipynb
+++ b/examples/extreme_response_full_sea_state_example.ipynb
@@ -464,7 +464,8 @@
     "This object provides common statistical functions (PDF, CDF, PPF, etc.) and metrics (expected value, median, etc). \n",
     "Here, we will look at the survival function and the 100-year return level. \n",
     "\n",
-    "The value of the survival function at a given return level (e.g. 100-years) (`s_t` in the code below) is $1/N$ where $N$ is the number of short-term periods in the return period. In this case $N$ is the number of 3-hour periods in a 100-year period, which gives `s_t`$\\approx 3e-6$. The corresponding response, i.e. the 100-year wave elevation in this case, is given as the inverse cumulative function (ppf) of $1-$`s_t`. This gives a 100-year wave of about 10.4 meters.  "
+    "The value of the survival function at a given return level (e.g. 100-years) \n",
+    "may be calculated using the `return_year_value` function. This gives a 100-year wave of about 11.1 meters.  "
    ]
   },
   {
@@ -483,8 +484,7 @@
    "source": [
     "t_st_hr = t_st/(60.0*60.0)\n",
     "t_return_yr = 100.0\n",
-    "s_t = 1.0/(365.25*24*t_return_yr/t_st_hr)\n",
-    "x_t = lte.ppf(1-s_t)\n",
+    "x_t = extreme.return_year_value(lte.ppf, t_return_yr, t_st_hr)\n",
     "\n",
     "print(f\"100-year elevation: {x_t} m\")"
    ]
@@ -493,7 +493,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Finally we plot the survival function and show the 100-year return level (dashed grey lines). The 100-year value is about 104 m (where the grey line intersects the x-axis)"
+    "Finally we plot the survival function and show the 100-year return level (dashed grey lines). The 100-year value is about 11.1 m (where the grey line intersects the x-axis)"
    ]
   },
   {
--- a/mhkit/loads/extreme.py
+++ b/mhkit/loads/extreme.py
@@ -727,3 +727,31 @@ def mler_export_time_series(rao, mler, sim, k):
     mler_ts = mler_ts.rename(columns={0: 'WaveHeight', 1: 'LinearResponse'})
 
     return mler_ts
+
+
+def return_year_value(ppf, return_year, short_term_period_hr):
+    """
+    Calculate the value from a given distribution corresponding to a particular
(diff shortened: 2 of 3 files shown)


mbruggs:
Thanks for looking @ssolson! Sorry for not adding the test, I completely missed the test directory when I looked. I'll add a test once we have some feedback from Carlos.

I've run autopep8 over the function and pushed the changes but let me know what other formatting changes are needed.

cmichelenstrofer:
@ssolson Yes this is very useful. I was doing it manually in the examples but having a function is better.

examples/extreme_response_full_sea_state_example.ipynb
```diff
@@ -493,7 +493,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Finally we plot the survival function and show the 100-year return level (dashed grey lines). The 100-year value is about 104 m (where the grey line intersects the x-axis)"
+    "Finally we plot the survival function and show the 100-year return level (dashed grey lines). The 100-year value is about 11.1 m (where the grey line intersects the x-axis)"
```
cmichelenstrofer, line 496: 104m lol :) Glad someone else looked at this example!

mhkit/loads/extreme.py
```diff
@@ -727,3 +727,31 @@ def mler_export_time_series(rao, mler, sim, k):
...
+        value: float
+            The value corresponding to the return period from the distribution.
+    """
+    assert callable(ppf)
+    assert isinstance(return_year, int)
+    assert isinstance(short_term_period_hr, (float, int))
+
+    p = 1 / (return_year * 365.25 * 24 / short_term_period_hr)
+
+    return ppf(1 - p)
```
cmichelenstrofer, line 757: Great!

cmichelenstrofer (approved):
Looks great, thanks for doing this!

akeeste:
@mbruggs Carlos approved your changes. Can you add a test for this function? There are plenty of examples in [/tests/test_loads/](https://github.com/MHKiT-Software/MHKiT-Python/blob/master/mhkit/tests/test_loads.py). I'm also going to change this PR to go into our Develop branch, per our current workflow. @ssolson or I can do a final review once the test is in. 

Adam

mbruggs:
Thanks for the input everyone, much appreciated. I've added a test but it found it hard to know exactly what the test should be. In the end, the test is against a 'known value', ie computed with the function under test itself. This should, at least, guard against any unexpected changes in the function signature or calculation, assuming that the current calculation is correct.

Sorry for missing that the branch should be merged into Develop. I've rebased my branch against Develop so the merge should be neat.

Thanks again for all the work on MHKiT!

mhkit/tests/loads/test_loads.py
```diff
@@ -180,6 +180,20 @@ def test_mler_export_time_series(self):
...
+        dist = stats.norm
+        return_year = 50
+        short_term_period = 1
+
+        val = loads.extreme.return_year_value(dist.ppf, return_year, short_term_period)
+
+        # This is a little hard to test as there is not an indepedent means by
+        # which we can do this calculation. The value below has been taken from
+        # an initial calculation and will indicate if the return_year_value is
+        # changed unexpectedly.
```
akeeste, line 193: I think this is acceptable right now, I can think of a couple other tests that do something similar. @ssolson or @cmichelenstrofer thoughts on alternative known values?

Can you remove the comment lines, those will be recorded in this thread.

mhkit/loads/extreme.py
```diff
@@ -727,3 +727,31 @@ def mler_export_time_series(rao, mler, sim, k):
...
+def return_year_value(ppf, return_year, short_term_period_hr):
+    """
+    Calculate the value from a given distribution corresponding to a particular
+    return year.
+
+    Parameters
+    ----------
+    ppf: callable function of 1 argument
+        Percentage Point Function (inverse CDF) of short term distribution.
+    return_year: int
```
akeeste, line 741: Does ``return_year`` have to be an int? I think we could allow floats here as well without issue.

akeeste (changes requested):
@mbruggs I made a few minor clean-up comments. Otherwise the function is appropriately short and clean. I don't have much else, aside from getting a second opinion from @cmichelenstrofer or @ssolson on what the test is against.

mhkit/tests/loads/test_loads.py
cmichelenstrofer: This is a good enough test.

mbruggs: Thanks both for confirming.

mhkit/loads/extreme.py
mbruggs: Good point, I agree that it could be a float. Updated.

PR #200, opened 2022-11-07
Run CI on push to Develop

ssolson:
Currently, we are not running tests when pushing to the Develop branch.

This will enable CI tests on push to the Develop branch


akeeste (approved):
@ssolson Good idea. This will be helpful when going through our current PR list. It seems like each PR needs the most recent version of ``Develop`` pulled in again before the updated test suite will run. This can help us confirm that each PR runs as expected without needing to update each branch

Issue #204, opened 2022-11-14
Wave annual average energy matrix graphic bug


alimayt:
Hello. I found an error in the calculation performed in the graphics.py for plot_avg_annual_energy_matrix. The code should be taking an average (I believe), but instead it is a loop that iteratively adds the next value to the previous value and divides by 2, meaning (for example), for 3 years the loop would give an output of x0/4 + x1/4 +x2/2 instead of x0/3+ x1/3 + x2/3. 

Beginning at line 455, it reads, 
    # Initialize average annually with first year
    avg_annual_counts_hist = hist_counts[years[0]]
    avg_annual_J_hist = hist_J[years[0]]

    # Iterate over each year and average
    for year in years[1:]:
        avg_annual_counts_hist = (avg_annual_counts_hist+ hist_counts[year])/2
        avg_annual_J_hist = (avg_annual_J_hist+ hist_J[year])/2

When I think it should read:
    # Initialize average annually with first year
    avg_annual_counts_hist = hist_counts[years[0]]/len(years)
    avg_annual_J_hist = hist_J[years[0]]/len(years)

    # Iterate over each year and average
    for year in years[1:]:
        avg_annual_counts_hist = avg_annual_counts_hist+ hist_counts[year]/len(years)
        avg_annual_J_hist = avg_annual_J_hist+ hist_J[year]/len(years)

cmichelenstrofer:
current:
```python
# Initialize average annually with first year
avg_annual_counts_hist = hist_counts[years[0]]
avg_annual_J_hist = hist_J[years[0]]

# Iterate over each year and average
for year in years[1:]:
    avg_annual_counts_hist = (avg_annual_counts_hist+ hist_counts[year])/2
    avg_annual_J_hist = (avg_annual_J_hist+ hist_J[year])/2
```

Proposed:
```python
# Initialize average annually with first year
avg_annual_counts_hist = hist_counts[years[0]]/len(years)
avg_annual_J_hist = hist_J[years[0]]/len(years)

# Iterate over each year and average
for year in years[1:]:
    avg_annual_counts_hist = avg_annual_counts_hist+ hist_counts[year]/len(years)
    avg_annual_J_hist = avg_annual_J_hist+ hist_J[year]/len(years)
```

cmichelenstrofer:
@ssolson @rpauly18 you wrote this function originally, can you confirm it is a bug?

ssolson:
[...]

PR #205, opened 2022-11-16
Bug Fix: Averaging histogram bin and wave energy

ssolson:
Addresses issue #204 plus minor formatting updates to `wave/graphics.py`


ssolson:
@cmichelenstrofer would you mind taking a look at this PR? @akeeste is OOO.

cmichelenstrofer (approved):
This PR 

- applies the suggested changes in #204
- changed bin count border and color
- several small formatting changes.

I didn't actually try to understand the code, but since both @ssolson  and @alimayt agree this is the correct solution, I trust it is right now :)

cmichelenstrofer:
@ssolson I can't think of any way of adding a test since the output is a figure. How can we catch this type of things in the future?

ssolson:
> @ssolson I can't think of any way of adding a test since the output is a figure. How can we catch this type of things in the future?
Off the top of my head, we would separate out the calculation of the QoIs and the plotting into multiple functions. Given the other PRs/Issues I would rank this as a lower priority but something to consider as a "Good first Issue".

PR #196, opened 2022-10-24
Fix: update variable name to remove reference before assignment error

Graham-EGI:
Snippet from error (happens a bit after the changed line in PR): 
```
 y = y1[ms]
UnboundLocalError: local variable 'y1' referenced before assignment
```

PR #208, opened 2022-11-29
Fix: update variable name to remove reference before assignment error

ssolson:
Supersedes #196 by @Graham-EGI.

> Snippet from error 
```
 y = y1[ms]
UnboundLocalError: local variable 'y1' referenced before assignment
```


ssolson:
Reviews & discussion in #196. Merging into `Develop` to Fix Bug.

PR #211, opened 2022-12-08
HSDS

ssolson:
This PR will fix the issues with the hindcast direction wave spectrum by breaking the data request into multiple requests to the API. As such the hindcast tests are added back into the CI testing suite.


ssolson:
The final steps to fix the issue I took were:
1. Only windows and max
2. Only python 3.9
3. Add a retry with an exponential backoff wait time

mhkit/wave/io/hindcast.py
```diff
@@ -216,13 +218,69 @@ def request_wpto_directional_spectrum(lat_lon, year, tree=None,
...
+                try:
+                    data_array = rex_waves[parameter, bins[i]:bins[i+1], :, :, gid]
+                    str_error = None
+
+                except Exception as e:
+                    str_error = str(e)
+
+                if str_error:
+                    sleep(sleep_time)
+                    sleep_time *= 2
```
akeeste, line 264: Let's set a maximum ``sleep_time`` to prevent this from getting very large. Unless something else will break the job at those long waits, we'll take up a lot run time on GitHub Actions virtual machines

mhkit/wave/io/hindcast.py
```diff
@@ -216,13 +218,69 @@ def request_wpto_directional_spectrum(lat_lon, year, tree=None,
...
+        
+            ax1 = np.product(data_array.shape[:3])
+            ax2 = data_array.shape[-1] if len(data_array.shape) == 4 else 1
+            data_array = data_array.reshape(ax1, ax2)        
+            
+            df = pd.DataFrame(data_array, columns=columns, index=idx)
+            df.name = parameter
+            datas[i]=df
+
+        # Appednd each request into an xarray
```
akeeste, line 276:
```suggestion
        # Append each request into an xarray
```

mhkit/wave/io/hindcast.py
```diff
@@ -216,13 +218,69 @@ def request_wpto_directional_spectrum(lat_lon, year, tree=None,
...
+        
+        # Request multiple datasets and add to dictionary
+        datas={}
+        for i in range(len(bins)-1):
+            idx=index[index_bins[i]:index_bins[i+1]] 
+
+            # Request with exponential back off wait time
+            sleep_time = 2
+            num_retries = 4
+            for x in range(0, num_retries):
```
ssolson, line 254: @akeeste the sleep time should only increase by the number of retries which is currently 4. Which would lead to a total max time of 2^4 seconds of wait time. Does this address your comment?

mhkit/wave/io/hindcast.py
ssolson: The max sleep time should be dictated by the max number of retries. Please see my comment below.

mhkit/wave/io/hindcast.py
akeeste: @ssolson yes that does, thanks. Missed that on my first read

PR #224, opened 2023-02-16
Update readme

castillocesar:
draft of updated readme. still need to update the pull request instructions.


castillocesar:
updated the README to include some instructions on submitting issues, creating a fork, creating a branch, and submitting a pull request.

ssolson:
Hey @castillocesar thanks for this submission. Looking at this PR you have added 4 sections to the MHKiT README that cover the basic features of GitHub. I would currently not recommend this PR as written for inclusion into MHKiT. At most I would only include how to create an issue at which point the maintainers could help interested parties lacking version control experience. Further much of this is covered in detail with not only the how but the why in the documentation.

@akeeste and @rpauly18 could you provide your thoughts?

rpauly18:
@ssolson I see what you are saying about it being repeated information from the documentation, but I do not think it hurts anything to have it repeated here. Hopefully this will be a reminder to people of how to operate when contributing to MHKiT. If you feel strongly that this should not be included, I won't get upset about that either

PR #225, opened 2023-02-22
Fix type for return_year in extreme.return_year_value

mbruggs:
In an oversight in merging #193 the docstring was updated for the type of return_year in return_year_value but the corresponding assertion was not.

Tests for the types of the two input parameters have been added to try to avoid a simular situation occuring the future.

### DIFF
--- a/mhkit/loads/extreme.py
+++ b/mhkit/loads/extreme.py
@@ -749,7 +749,7 @@ def return_year_value(ppf, return_year, short_term_period_hr):
         The value corresponding to the return period from the distribution.
     """
     assert callable(ppf)
-    assert isinstance(return_year, int)
+    assert isinstance(return_year, (float, int))
     assert isinstance(short_term_period_hr, (float, int))
 
     p = 1 / (return_year * 365.25 * 24 / short_term_period_hr)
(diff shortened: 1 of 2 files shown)


mbruggs:
Sorry about this, my mistake in the initial PR. I found it when trying to integrate v0.6.0 into a project. Is there a 'hot-fix' procedure that would mean a v0.6.1 would be possible with just this change (or any others that fix bugs)?

mbruggs:
Looks like there are quite a few test failures but they don't appear to be related to this code change.

ssolson:
No worries Mark we can tag a new release with this change. 

I am looking into the other failures. At first glance, It looks like matplotlib updated.

mbruggs:
Thanks @ssolson!

ssolson:
#226 should fix the issues with the matplotlib update

ssolson:
@mbruggs it is showing that you are trying to change 190 files currently. 

Can you resubmit this PR using the latest version of `develop` as your HEAD so that we are sure we are only changing what we intend to?

ssolson:
Okay @mbruggs I fixed the issues causing the tests to fail by merging #226 . If you pull the latest `develop` branch everything should pass.

mbruggs:
Thanks for fixing the matplotlib issue so quickly, @ssolson. It seems to have fixed the tests except the hindcast ones which seem to be hanging (I couldn't find a way to stop them).

Branch has been rebased so now only showing the files that have actually been changed.

ssolson (approved):
This looks great. Thank you for this PR!

PR #235, opened 2023-04-13
Mooring module

hivanov-nrel:
# Mooring module

create functions for mooring module:

* [x] read MoorDyn OUT file 
* [x] read MoorDyn input file
* [x] calculate lay length
* [x] create 3d animation
* [x] create 2d animation

## Tests created for
* [x] io
* [x] general
* [x] graphics


hivanov-nrel:
@ssolson @rpauly18 I think a first pass at the mooring IO module is complete at this point. It's one function but it can ingest both the input and output files from Moordyn. There is still room for refinement and added functionality, but I need to find more complicated example output files than what's currently available to me. 

I'm working on a couple of simple functions that then utilize the output data to do some calculations and plotting. At this point, my question is, should we try to squeeze these extra functions into this PR or should we separate them into a later PR? What are your thoughts?

rpauly18:
@hivanov-nrel keep working off of this PR so we can have a complete example to work from on the review.

hivanov-nrel:
@ssolson @rpauly18 okay I think this is ready for review. Not sure who wants to take point on that. 

I definitely know of some future functionality we could add, but in the interest of time, I thought this would be okay to start. Please let me know if you disagree.

rpauly18:
@hivanov-nrel I agree that this looks like a good start. We will figure out who has time to tackle this review and make a formal assignment. Thanks!

mhkit/mooring/general.py
```diff
@@ -0,0 +1,54 @@
+import xarray as xr
+
+
+def laylength(ds, depth, tolerance=0.25):
```
ssolson, line 4: MHKiT uses `snake_case` everywhere

mhkit/mooring/general.py
```diff
@@ -0,0 +1,54 @@
...
+    """Calculate the laylength of a mooring line over time.
```
ssolson, line 5: MHKiT starts docstring contents on a new line

SHould be:
```
"""
Calculate the laylength of a mooring line over time.
...
```
Same everywhere

mhkit/mooring/general.py
```diff
@@ -0,0 +1,54 @@
...
+        Array containing the laylength at each time step.
+
+    Raises
+    ------
+    ValueError
+        Checks for mininum number of nodes necessary to calculate laylength.
+    """
+    assert isinstance(ds, xr.Dataset), 'ds must be of type xr.Dataset'
+    assert isinstance(depth, (float,int)), 'depth must be of type float or int'
+    assert isinstance(tolerance, (float,int)), 'tolerance must be of type float or int'
```
ssolson, line 28: Going forward we are no longer using assert statements in functions (see #242). In your case these would no be

```
    if not isinstance(ds, xr.Dataset):
        raise TypeError('ds must be of type xr.Dataset')
    if not isinstance(depth, (float, int)):
        raise TypeError('depth must be of type float or int')
    if not isinstance(tolerance, (float, int)):
        raise TypeError('tolerance must be of type float or int')
```

mhkit/mooring/general.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Returns
+    -------
+    xr.Dataset
+        Array containing the laylength at each time step.
+
+    Raises
+    ------
+    ValueError
+        Checks for mininum number of nodes necessary to calculate laylength.
```
ssolson, line 24: This is a great addition. Based on the change below I think would be updated to:

```
Raises
------
ValueError
    Checks for minimum number of nodes necessary to calculate laylength.
TypeError
    Checks for correct input types for ds, depth, and tolerance.
    ```

**Review by ssolson — COMMENTED** [review]
@hivanov-nrel thanks for getting this started. I have made some general comments on this first function. Could you propagate these suggestions through the PR?

**Comment by hivanov-nrel** [comment]
@ssolson Sorry for the holdup; i just pushed a commit that addresses your comments. Please let me know if there is anything else we need. Have you gone through the example notebook?

**Comment by ssolson** [comment]
> @ssolson Sorry for the holdup; i just pushed a commit that addresses your comments. Please let me know if there is anything else we need. Have you gone through the example notebook?

Thanks @hivanov-nrel . My last comments were simply a quick check to knock out some easy fixes. I am going to take a deeper look today. I will go run the example notebook now.

**Comment by ssolson** [comment]
@hivanov-nrel some suggested updates are in a PR on your fork

https://github.com/hivanov-nrel/MHKiT-Python/pull/6

**Comment by hivanov-nrel** [comment]
@ssolson thanks for your suggestions. I've approved the PR and merged into my fork. Let me know if there any other changes we need to make to this PR before merging to the dev branch

**Review by ssolson — COMMENTED** [review]
Chris checkout my PR for my needed changes.

I just have 1 question below and the issue opened and I we are good to go on this PR.

File: mhkit/mooring/general.py [hunk 5]
```diff
@@ -0,0 +1,57 @@
...
+    lay_z = nodesZ[lay_indx]
+    # get corresponding x-point and y-point node names 
+    lay_x = lay_z[:-1] + 'x'
+    lay_y = lay_z[:-1] + 'y'
+    lay_0x = nodesX[0]
+    lay_0y = nodesY[0]
+    # find distance between initial point and lay point
+    laylength_x = ds[lay_x] - ds[lay_0x]
+    laylength_y = ds[lay_y] - ds[lay_0y]
+    lay_length =  (laylength_x**2 + laylength_y**2) ** (1/2)
```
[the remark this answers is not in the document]
hivanov-nrel: Technically, the function is actually already making this calculation in 3D. By definition of the lay length of a mooring line, the z-component of this length will always be zero and therefore will not contribute when doing the calculation. Let me know if that doesnt make sense.

ssolson (approved):
Approved for merge into development.

The PR failed 1 of 2 hindcast tests but that is unrelated to this PR.

PR #240, opened 2023-05-11
Package Requirements & Python version 10, 11

ssolson:
This PR adds testing support for Python 3.10 and 3.11 and removes all upper bound on required package versions such that users can work with the newest versions of all required packages.

Additionally this PR will now only run on ubuntu on pulls to `develop` and will run windows, macOS, and ubuntu on pulls to `master`. This reduces runtime by ~15 minutes for development PRs. To make this work a new job was added `set-os` which sets the os matrix based on which branch the PR is made against.

Most changes did not require changing the module source code. To update to the latest pandas:
- `mhkit.power.quality`: change deprecated `get_loc` to `get_indexer` ([docs](https://pandas.pydata.org/docs/reference/api/pandas.Index.get_indexer.html))

image


ssolson:
Pecos has merged a fix to master causing the error seen in the current build. Specifically the error is from pandas index method deprecation which replaced `is_montonic` with `is_monotinic_increasing` (or decreasing) and will soon release an update to pypi which will allow MHKiT to pass tests with the most recent pandas version.

https://github.com/sandialabs/pecos/issues/76

ssolson:
@akeeste this PR is ready for review.

One discussion point I would suggest adding would be to only test ubuntu on pulls to dev and then test windows and mac on pull to master. Doing so would reduce testing time from ~30 minutes to ~15 minutes.

image

akeeste (changes requested):
@ssolson Thanks for these updates!

Multiple OS increasing the test time has recently been an issue for WEC-Sim too. I support only running Mac and Windows for changes to master. It's worth a try to speed-up development time.

A couple requests:
- Update the job name to reflect the new python versions being test
- Update the ``hindcast-calls`` jobs to mirror either the ``pip-build`` or ``conda-build``? Now that the ``check-changes`` and ``prepare-cache`` jobs are in place, it shouldn't add a lot of test time to keep it consistent with the other main jobs. 

I'm open to discussion here on the hindcasts since it's been such a big issue lately

ssolson:
@akeeste I believe I have addressed your changes as well as added the only ubuntu on pulls to `develop`. Please, give it a review and lmk what you think.

PR #241, opened 2023-05-11
Improve Testing Time

ssolson:
The primary purpose of this PR is to decrease the testing time. This will speed up all other developments. The primary approach here is to use data caching to only call the APIs once. This should additoinally reduce test failures and subsequent test re-runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,15 +37,12 @@ jobs:
           export PATH="${CONDA_PREFIX}/bin:${CONDA_PREFIX}/Library/bin:$PATH" # so setup.py finds nc-config
           pip install -e . --no-deps --force-reinstall
 
-      - name: Tests
+      - name: Install dependencies
         shell: bash -l {0}
         run: |
           source activate TEST
           python -m pip install --upgrade pip wheel
-          pip install coverage
-          pip install pytest
-          pip install coveralls
-          pip install .
+          pip install coverage pytest coveralls .
 
       - name: Run pytest
         shell: bash -l {0}
@@ -86,10 +83,7 @@ jobs:
       - name: Update and install packages
         run: |
           python -m pip install --upgrade pip wheel
-          pip install coverage
-          pip install pytest
-          pip install coveralls
-          pip install .
+          pip install coverage pytest coveralls .
 
       - name: Run pytest
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,8 @@ jobs:
       matrix:
         os: ["windows-latest", "ubuntu-latest", "macos-latest"]
         python-version: [3.7, 3.8, 3.9]
+    env:
+      PYTHON_VER: ${{ matrix.python-version }}
 
     steps:
       - uses: actions/checkout@v2
@@ -32,7 +34,7 @@ jobs:
       - name: Python ${{ matrix.python-version }}
         shell: bash -l {0}
         run: |
-          conda create --name TEST python=${{ matrix.python-version }} numpy cython pip hdf5 libnetcdf cftime netcdf4 --strict-channel-priority
+          conda create --name TEST python=${PYTHON_VER} numpy cython pip hdf5 libnetcdf cftime netcdf4 --strict-channel-priority
           source activate TEST
           export PATH="${CONDA_PREFIX}/bin:${CONDA_PREFIX}/Library/bin:$PATH" # so setup.py finds nc-config
           pip install -e . --no-deps --force-reinstall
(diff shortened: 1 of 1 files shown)


ssolson:
API Calls to add cacheing to:
1. wave/
    - [ ]  ndbc
    - [ ] cdip
    - [ ] hidcast
2. tidal/
    - [ ]  noaa
3. river/
    - [ ] usgs

ssolson:
@jmcvey3 I have not looked at your latest PR or into these error but have you already solved these dolfyn xarray failures?
image

ssolson:
## Wind Tool API issues

 This post aggregates useful resources and communicates my current progress.

My current road block on this PR is the wind-hindcast bc the API will not consistiently return data. Similar to what I did for the wave hindcast I have been digging into the base layer of the API calls using bothNREL-REX and h5pyd.


### Wind Data Endpoints

To start a list of available data can be found here:
> https://github.com/NREL/hsds-examples/blob/master/datasets/WINDToolkit.md

E.g. an endpoint we could use:
`/nrel/wtk/offshore_ca/Offshore_CA_2000.h5`


### Examples of accessing the endpoints

Exampls on how to access the data at these endpoints are covered in these notebooks:

> 1. https://github.com/NREL/hsds-examples/blob/master/notebooks/01_WTK_introduction.ipynb
> 2. https://github.com/NREL/hsds-examples/blob/master/notebooks/08_NREL-rex.ipynb


### Example using h5pyd
So using h5pyd the minimumworking example would be:
```
import h5pyd
data_path = '/nrel/wtk/offshore_ca/Offshore_CA_2000.h5'
f = h5pyd.File(data_path, 'r')
dset = f['windspeed_100m']
dset[1, 1]
```
This consistiently returns 
>  *** ValueError: cannot reshape array of size 0 into shape (1,)


### Example using NREL-REX
We can do the same using NREL-REX as follows:
```
from rex import WindX
data_path = '/nrel/wtk/offshore_ca/Offshore_CA_2000.h5'
with WindX(data_path , hsds=True) as rex_wind:
    ds = rex_wind['windspeed_100m', 1, 1]
```
Which again will constiently return an array of size 0
> ValueError: cannot reshape array of size 0 into shape (1,)

ssolson:
Raised the issue with the NREL development folks

https://github.com/NREL/developer.nrel.gov/issues/320

ssolson:
Looking at the prepare cache vs hindcast we can see that the hindcast test only took 2 minutes.
image

ssolson:
Something is going on with python 3.9 and the test_kde_copula of the contours. I have not modified this file so most likely a dependency change. The test completes using 3.9 on my local machine without error.
image

ssolson:
Matplotlib (https://github.com/matplotlib/matplotlib) release v3.8.0 5 days ago and this deprecated the following lines (1554:1556) in contours.py:
```python 
    for i, seg in enumerate(vals.allsegs[0]):
        x1_bivariate_KDE.append(seg[:, 1])
        x2_bivariate_KDE.append(seg[:, 0])
```        

Specifically `vals` is a variable of type QuadContourSet. and the updated call should be:
```python
for i, seg in enumerate(vals.get_paths()):
```

ssolson:
😃 oc the fix is not backwards compatible with 3.8
image

ssolson:
Python 3.7 is failing for all tests due to a Dolfyn method and an xarray new release. 
- In our current tests on master we only run py 3.7 tests on a conda install (no pip install or hindcast checks).  
- MHKiT's latest master release was on August 11.  
- xarray released a new version on August 19 that seems to have broken this the py 3.7 conda compatibility.  

Python 3.7 was released on 27 June 2018 making it over 5 years old. My vote is to drop 3.7 support in this PR and in a follow on PR add Python 3.10 support for the next MHKiT release.

image

mhkit/wave/io/hindcast/wind_toolkit.py
ssolson, line 1: - add caching
- breakout get region into own function
- add file docstring
- assert -> exceptions
- lint

mhkit/wave/io/ndbc.py
ssolson, line 1: - add caching
- requests are in try, except pattern
- lint

setup.py
ssolson, line 1: - lint

akeeste (commented):
@ssolson I added a couple minor docstring revisions on my first pass. Thanks for documenting what's going on and where, that was helpful. I want to give this another pass and make sure I'm understanding everything. I'll add any more questions/comments then.

akeeste (approved):
@ssolson Overall the caching seems clean, well documented, and functional. The decreased time is a big improvement. I'm not familiar with the caching process beyond this review though. So I'll defer to your judgement on the implementation and we can learn as we go if issues arise.

Issue #229, opened 2023-03-15
Improve execution time for surface elevation computation


mbruggs:
This is primarily placeholder, I'm happy to contribute this improvement when I get a moment. The code below needs generalising to properly use the Dataframe that is passed in.

### Describe the bug:
The computation of the surface elevation from a spectrum uses a 'sum of sines' approach which is prohibitively slow. During some experiments, trying calculate a 3hr sea state at 20Hz resolution crashed the python process on my machine, a 1hr sea state took 11s.

### Alternative approach
Alternatively, we can use an approach using an `ifft`. For the same calculation that previously took 11s, the `ifft` version took 0.007s.

The code below (loosely adapted to match current MHKit structure) gave the same results as the previous surface_elevation method using the sum of sines.

```python
def surface_elevation(S, time_index, seed=None, frequency_bins=None, phases=None)
    duration_s = time_index[-1]
    dt = time_index[1] - time_index[0]

    # see notes section in
    # https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html
    if time.size % 2 != 0:
        time = np.append(time, [duration_s + dt])

    f = S.index
    df = f[1] - f[0]
    if phases is None:
        np.random.seed(seed)
        phases = 2 * np.pi * np.random.rand(f.size)

    # Convert from energy to amplitude spectrum
    amp = np.sqrt(2 * S.values.squeeze() * df)
    amp[0] = 0  # 0 frequency gives NaN value in spectrum

    amp_re = amp * np.cos(phases)
    amp_im = amp * np.sin(phases)
    amp_complex = amp_re + 1j * amp_im

    eta = np.fft.irfft(0.5 * amp_complex * time.size, time.size)
```

ssolson:
That is an incredible time improvement and we should use this assuming there are no other tradeoffs. @cmichelenstrofer could you comment on @mbruggs proposed approach?

ryancoe:
I'd say it'd be relatively easy to check that frequency spacing and then select either the `ifft` or summation of sinusoids depending on whether the spacing is equal.

PR #250, opened 2023-07-13
Compute surface elevation using IFFT

mbruggs:
As discussed in #229 computing the surface elevation using the 'sum of sines' method is slow. Instead, we can use an IFFT which drastically increases performance. The IFFT method is used as the default when no frquency bins are provided and the frequency vector is equally spaced.

To compute the IFFT the zero frequency must be defined. The spectrum methods have been updated to avoid NaN values in the presence of the zero frequency. This means that spectrum and surface elevations functions work seamlessly together.


ssolson:
@mbruggs you da man. 

Looking this over I think we can get this in our next release for mid-August. I promise to get you a review before then. Currently, I am reviewing several other PRs for the next release so my review of this may be a week or so out.

Thank you so much for this contribution!

ssolson:
@akeeste could you review Mark's application of the IFFT? I am looking for help reviewing the application of the technique as I am not familiar with it.

akeeste:
@ssolson Yes definitely. @mbruggs I'll review this PR next week

akeeste (changes requested):
Hi @mbruggs 

Thanks for this great speed update. I only have a few changes to make, which I put in mbruggs/MHKiT-Python#1, including
- [x] add doc string for frequency_bins
- [x] update example notebook calls to surface_elevation to use 0 frequency and the faster IFFT method

Once that is merged, I'll go ahead and bring this into MHKiT

Issue #242, opened 2023-05-19
Assert Statements are for Debugging


ssolson:
Currently MHKiT uses assert staements in its production code (primarilty for type checking functional inputs). As stated in the docs (https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) assert stements are for debugging. 

We should raise an error if the user does not pass the correct information. 

So going forward we would convert:
```python
assert isinstance(station_number, (str, type(None))), (f'station_number must be  of type string')
```
to this

```python
if not isinstance(station_number, (str, type(None))):
    raise ValueError(f'station_number must be of type string. Got: {station_number}')
```

ssolson:
Here is an article covering this idea in more detail:
https://snyk.io/blog/the-dangers-of-assert-in-python/

![image](https://github.com/MHKiT-Software/MHKiT-Python/assets/13438942/8affe511-ab42-4a41-8316-65e1a5dc36f9)

onedeeper:
Hello! I'm trying to get my feet wet with some python open source. This seems straight forward so I'm happy to take care of it. Does this seem like a good first issue? Thanks!

PR #255, opened 2023-08-09
add function cumulative rainflow spectrum

kchong75:
@rpauly18 @hivanov-nrel 
PR for mhkit-fatigue: cumulative rainflow spectrum

### DIFF
--- a/mhkit/loads/general.py
+++ b/mhkit/loads/general.py
@@ -165,3 +165,53 @@ def damage_equivalent_load(data_signal, m, bin_num=100, data_length=600):
     DEL = DELs.sum() ** (1/m)
 
     return DEL
+
+def cumulative_rainflow_spectrum(data_signal, m, bin_num=100, data_length=600):
+    '''
+    Calculates the cumulative rainflow spectrum for power production for each load quantity.
+    Summing all the individual rain flow cycle counts of each file in the power production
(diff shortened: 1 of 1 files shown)


mhkit/loads/general.py
```diff
@@ -165,3 +165,53 @@ def damage_equivalent_load(data_signal, m, bin_num=100, data_length=600):
...
+      techniques for accurate fatigue damage estimation. International Journal
+      of Fatigue, 82 (2016) 757-765`
+
+
+    Parameters:
+    -----------
+    data_signal : 2d array
+        Data signals being analyzed
+    bin_num : int
+        Number of bins for rainflow counting method (minimum=100)
```
rpauly18, line 190: I would add an assert statement checking to see if this value is above 100

ssolson: Hey @rpauly18 we are no longer using assert statements. See #242.

For the current assert the replacement would be:
```
    if not isinstance(bin_num, (float, int)):
        raise ValueError('bin_num must be of type float or int')
```

ChatGPT is pretty good at converting these but we need to always check its work.

rpauly18:
@hivanov-nrel can you take a look at this and make sure the functionality is as expected? Can you also develop a quick test for this?

PR #265, opened 2023-08-29
Pull develop doc fixes into master 

akeeste:


ssolson:
Deprecate for #266

Need to pull master into develop

PR #214, opened 2023-01-03
Automatic optimal threshold for peaks-over-threshold method

cmichelenstrofer:
Automatic threshold for peaks-over-threshold developed by Vince Neary and Seongho Ahn.

PR #268, opened 2023-10-03
Automatic threshold for peaks-over-threshold

cmichelenstrofer:
replaces #214.

MHKiT already has a peaks-over-threshold method, which takes in threshold as one of the inputs. This method here is to automatically calculate the "best" threshold. It is from a paper by Vince and Seongho:

Neary, V. S., S. Ahn, B. E. Seng, M. N. Allahdadi, T. Wang, Z. Yang and R. He (2020). "Characterization of Extreme Wave Conditions for Wave Energy Converter Design and Project Risk Assessment.” J. Mar. Sci. Eng. 2020, 8(4), 289; https://doi.org/10.3390/jmse8040289.

They had it in Matlab and I converted it to Python here.


cmichelenstrofer:
@ssolson I cleaned this up quite a bit and made a Matlab minimal example form the code I got from Seongho. He unfortunately left which makes figuring some of this out more difficult. The problem is that the solution I get with my Python implementation is slightly different from what I get with the original Matlab code. I think the issue might be that I am not accounting for all the metrics they used for "independent storms". I would appreciate some help if anyone else can take a look at this. 

The failing errors don't seem to be related to this PR.

ssolson:
> @ssolson I cleaned this up quite a bit and made a Matlab minimal example form the code I got from Seongho. He unfortunately left which makes figuring some of this out more difficult. The problem is that the solution I get with my Python implementation is slightly different from what I get with the original Matlab code. I think the issue might be that I am not accounting for all the metrics they used for "independent storms". I would appreciate some help if anyone else can take a look at this.
> 
> The failing errors don't seem to be related to this PR.

Carlos thank you. 

The errors are unrelated to this PR and are fixed in #241 which should be reviewed and approved soon.

I took a bit of a look at his today. Running the tests I see that I get the same answer as you have specified but I was wondering how big is the difference from the Matlab version?

I started looking into the generalized pareto distribution in scipy vs matlab via the docs but I was ultimately hindered by not being able to run the Matlab version without getting Matlab installed on my machine. It seems like the idea would be to step-by-step compare the python vs Matlab output to see where the difference arises.

@akeeste are you setup to run Matlab and do you have time to take a look into this further?

akeeste:
@ssolson I'm set-up for this. I will review this and debug next week.

cmichelenstrofer:
@ssolson @akeeste Thanks! The difference I'm getting is this:

Matlab: 99.53%, 1.041 m
Python: 98.94%, 1.028 m

@akeeste LMK if you want to go over it together.

akeeste:
@cmichelenstrofer I made some progress today and noting where I'm at for reference. I noticed a couple things, but I'm still trying to figure out how much they affect the end result, given the input Hs used in your working example. I'll refer to Seongho's original code as the 'matlab' version and your refactored code as the 'python version'
- xarray's ``stats.scoreatpercentile`` gives slightly different values than MATLAB's ``prctile`` function. The differences are very small but may become more significant as the threshold range is refined. Also, [they recommend ](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.scoreatpercentile.html)moving to ``numpy.percentile``
   - Percentile, xarray threshold, matlab threshold
   - 0.99, 1.0294, 1.0295
   - 0.991, 1.0311, 1.0313
   - 0.992, 1.0334, 1.0334
   - 0.995, 1.0406, 1.0506
- the refinement logic is working a little differently in your refactored version. the condition ``rate_per_year < 2`` is checked for every threshold and refinement iteration in Python, but only checked every refinement iteration in MATLAB. So all 6 initial thresholds are checked in MATLAB, but only the first 3 in Python.
- In python for the first refinement iteration, the thresholds 0.99, 0.991, 0.992 all give an identical correlation. This seems suspicious and is not the behavior seen in the MATLAB code

I'll continue debugging next week

akeeste:
I'm still digging into the settings of various functions, but I can confirm that the functions for percentile, Generalized Pareto distribution fitting, and probability plotting functions all give slightly different results between their respective MATLAB and SciPy versions. I'll see if there are optional settings we can use to converge the results

mhkit/loads/extreme.py
```diff
@@ -157,6 +158,129 @@ def peaks_distribution_weibull_tail_fit(x):
...
+    assert len(initial_threshold_range) == 3, (
+        'initial_threshold_range must be length 3')
+    assert isinstance(max_refinement, int)
+
+    range_min, range_max, range_step = initial_threshold_range
+    best_threshold = -1
+    years = len(peaks)/(365.25*24/sampling_rate)
+
+    def _peaks_over_threshold(peaks, threshold, sampling_rate):
+        threshold_unit = stats.scoreatpercentile(peaks, 100*threshold)
```
akeeste, line 214:
```suggestion
        threshold_unit = np.percentile(peaks, 100*threshold, method='hazen')

File: mhkit/loads/extreme.py [hunk 2]
```diff
@@ -157,6 +158,129 @@ def peaks_distribution_weibull_tail_fit(x):
...
+            range_min = thresholds[max_i - 1]
+            range_max = thresholds[max_i] + 5*range_step
+        elif max_i == 0:
+            range_min = thresholds[max_i] - 9*range_step
+            range_max = thresholds[max_i + 1]
+        else:
+            range_min = thresholds[max_i-1]
+            range_max = thresholds[max_i+1]
+
+    best_threshold_unit = stats.scoreatpercentile(peaks, 100*best_threshold)
```
akeeste, line 280:
```suggestion
    best_threshold_unit = np.percentile(peaks, 100*best_threshold, method='hazen')

File: mhkit/loads/extreme.py [hunk 3]
```diff
@@ -157,6 +158,129 @@ def peaks_distribution_weibull_tail_fit(x):
...
+        acf_positive = acf[(idx_zero):(idx_zero+nlags+1)] / acf[idx_zero]
+
+        window_size = sampling_rate * positive_lag[acf_positive<0.5][0]
+        # window size in "observations" instead of "hours" between peaks.
+        window = window_size / sampling_rate
+        # keep only independent storm peaks
+        for idx in idx_storm_peaks[1:]:
+            if (idx - idx_independent_storm_peaks[-1]) > window:
+                idx_independent_storm_peaks.append(idx)
+                independent_storm_peaks.append(peaks[idx])
```
akeeste, line 240: Without subtracting the threshold unit, independent_storm_peaks becomes a combination of the storm_peak (excess) and the Hs
```suggestion
                independent_storm_peaks.append(peaks[idx]-threshold_unit)

File: mhkit/loads/extreme.py [hunk 4]
```diff
@@ -157,6 +158,129 @@ def peaks_distribution_weibull_tail_fit(x):
...
+            elif peaks[idx] > independent_storm_peaks[-1]:
+                idx_independent_storm_peaks[-1] = idx
+                independent_storm_peaks[-1] = peaks[idx]
```
akeeste, line 243: Without subtracting the threshold unit, independent_storm_peaks becomes a combination of the storm_peak (excess) and the Hs
```suggestion
                independent_storm_peaks[-1] = peaks[idx]-threshold_unit

**Review by akeeste — COMMENTED** [review]
Suggestions to improve a couple minor issues with the storm peak calculation. Otherwise, the storm peak calculation seems to pull all of the same independent storm peaks that the original matlab code does

**Comment by cmichelenstrofer** [comment]
@akeeste this is great! Thanks for looking into this in so much detail! I think the results are basically identical. The very tiny difference is from the techniques to fit the distribution. Do you think it is ready merge?

**Comment by akeeste** [comment]
@cmichelenstrofer Sounds good. Can you first remove the MATLAB working example and updated the test result to 99.13% and 1.032m. After that we can merge

File: mhkit/tests/loads/test_loads.py [hunk 5]
```diff
@@ -218,6 +218,13 @@ def test_shortterm_extreme(self):
             ste = loads.extreme.ste(t, data, t_st, method)
             assert_allclose(ste.cdf(x), cdf_1)
 
+    def test_automatic_threshold(self):
+        filename = "data_loads_hs.csv"
+        data = np.loadtxt(os.path.join(datadir, filename), delimiter=",")
+        years = 2.97
+        pct, threshold = loads.extreme.automatic_hs_threshold(data, years)
+        assert np.isclose(pct, 99.13)
```
akeeste, line 226: @cmichelenstrofer Apologies I was unclear when I typed my last message. The percentile is still returned as a fraction, so we should compare to 0.9913

```suggestion
        assert np.isclose(pct, 0.9913)

**Comment by akeeste** [comment]
Seems like I could commit the suggestion directly to the branch. I mistyped last time, the percentile still returns as a fraction. I updated the test to compare to 0.9913. The tests should pass now

File: mhkit/tests/loads/test_loads.py [hunk 6]
```diff
@@ -218,6 +218,13 @@ def test_shortterm_extreme(self):
             ste = loads.extreme.ste(t, data, t_st, method)
             assert_allclose(ste.cdf(x), cdf_1)
 
+    def test_automatic_threshold(self):
+        filename = "data_loads_hs.csv"
+        data = np.loadtxt(os.path.join(datadir, filename), delimiter=",")
+        years = 2.97
+        pct, threshold = loads.extreme.automatic_hs_threshold(data, years)
+        assert np.isclose(pct, 0.9913)
+        assert np.isclose(threshold, 1.032)
```
akeeste, line 227: Updating test value one more time
```suggestion
        assert np.isclose(threshold, 1.032092)

File: mhkit/loads/extreme.py [hunk 7]
```diff
@@ -157,6 +158,129 @@ def peaks_distribution_weibull_tail_fit(x):
...
+    -------
+    best_threshold: float
+        Threshold that results in the best correlation.
+    """
+    assert isinstance(sampling_rate, (float, int)), (
+        'sampling_rate must be of type float')
+    assert isinstance(peaks, np.ndarray), 'peaks must be of type np.ndarray'
+    assert len(initial_threshold_range) == 3, (
+        'initial_threshold_range must be length 3')
+    assert isinstance(max_refinement, int)
```
ssolson, line 207: Asserts should be replaced with exceptions

ssolson:
```
    if not isinstance(sampling_rate, (float, int)):
        raise TypeError('sampling_rate must be of type float or int')
    
    if not isinstance(peaks, np.ndarray):
        raise TypeError('peaks must be of type np.ndarray')
    
    if not len(initial_threshold_range) == 3:
        raise ValueError('initial_threshold_range must be length 3')
    
    if not isinstance(max_refinement, int):
        raise TypeError('max_refinement must be of type int')
```

cmichelenstrofer: @ssolson  Good to see this change in MHKiT! I changed it to what you suggested.

ssolson:
Overall, I'm happy with this PR as. I have a few suggestions, neither required but I think would round out the PR nicely.

1. Consider consolidating all the `scipy` imports onto a single line.
2. I recommend adding a docstring at the beginning of the file. While we haven't formally adopted this in MHKiT's style guidelines, it's something I've been introducing in files I've worked on.

Lastly, I noticed the use of a nested function. Historically, we've kept these functions separate. While nesting does highlight that the inner function is only needed by the parent function, it can also make the code appear more cluttered. I'm not opposed to the current structure, but I wanted to initiate a brief discussion on this topic. What are your thoughts?

akeeste:
To wrap up this PR I 
- [x] broke out the nested function 
- [x] consolidated the scipy imports

I left the submodule-level docstring for another PR. All tests have passed. Merging this PR

PR #270, opened 2023-10-12
Run hindcast tests only on change to hindcast module

ssolson:
This Pull Request introduces a mechanism to selectively trigger the execution of expensive hindcast tests based on the context of the changes made. Specifically, the hindcast tests will only run under two scenarios:

1. If there are changes made to the modules within the `mhkit/wave/io/hindcast` directory.
2. If the merge request is targeted at the `master` branch.

To achieve this, a new job named `check-changes` has been added to the workflow, which precedes the `prepare-cache` job.

image

Here's how the check-changes job operates:

1. Detecting Module Changes:
    - Check if any files within the `mhkit/wave/io/hindcast` directory have been altered.
      - If yes, set `wave_io_hindcast_changed` to true.
      - If no, set `wave_io_hindcast_changed` to false.
2. Determining Hindcast Test Execution:
    - Check if the PR is against the master branch or if `wave_io_hindcast_changed` is true.
    - If either condition is met, set `should-run-hindcast` to true.
      - Otherwise, set `should-run-hindcast` to false.

Some additional minor modifications have been made to ensure the seamless integration of Coveralls reporting, regardless of whether the hindcast tests are executed or skipped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -241,18 +241,6 @@ jobs:
           COVERALLS_FLAG_NAME: ${{ matrix.python-version }}
           COVERALLS_PARALLEL: true
 
-  log:
-    needs: [conda-build, pip-build, hindcast-calls]
-    runs-on: ubuntu-latest
-    container: python:3-slim
-    steps:
-      - name: Log job results
-        run: |
-          always() &&
-          echo conda-build result: ${{ needs.conda-build.result }}
-          echo pip-build result: ${{ needs.pip-build.result }}
-          echo hindcast-calls result: ${{ needs.hindcast-calls.result }}
-
   coveralls:
     name: Indicate completion to coveralls.io
     needs: [conda-build, pip-build, hindcast-calls]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,6 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       wave_io_hindcast_changed: ${{ steps.changes.outputs.wave_io_hindcast }}
+      should-run-hindcast: ${{ steps.hindcast-logic.outputs.should-run-hindcast }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


ssolson:
@akeeste can I request a quick review on this PR to help us move PRs through review quicker due to less time spent on hindcast calls?

.github/workflows/main.yml
```diff
@@ -211,6 +244,7 @@ jobs:
   coveralls:
     name: Indicate completion to coveralls.io
     needs: [conda-build, pip-build, hindcast-calls]
+    if: always() && (needs.hindcast-calls.result == 'success' || needs.hindcast-calls.result == 'skipped')
```
akeeste, line 247: Is this indicating that if the hindcast tests are run but fail, coveralls reporting is skipped? Will coveralls still consider the hindcast module to be covered if the hindcast tests don't need to run?

.github/workflows/main.yml
```diff
@@ -41,11 +67,16 @@ jobs:
...
       - name: Prepare data
         run: |
-          # pytest tests/test_specific_file.py::TestClass::test_function
           source activate TEST
           pytest mhkit/tests/river/test_io.py
           pytest mhkit/tests/tidal/test_io.py
           pytest mhkit/tests/wave/io/test_cdip.py
+
+      - name: Prepare Hindcast data
+        if: (github.event.pull_request.base.ref == 'master' || needs.check-changes.outputs.wave_io_hindcast_changed == 'true')
```
akeeste, line 76: We should use the should-run-hindcast variable so that the test logic is not duplicated

```suggestion
        if: (needs.check-changes.outputs.should-run-hindcast == 'true')

akeeste (commented):
Hi Sterling, thanks for doing this. A couple minor comments included

.github/workflows/main.yml
ssolson: If hindcast fails coveralls is skipped. You can see that as the current behavior:
image


In my screenshot for this PR description I show the used logic. You can see that hindcast is skipped and that coveralls is submitted.

.github/workflows/main.yml
ssolson: Good call! Fixed in [b81f61f](https://github.com/MHKiT-Software/MHKiT-Python/pull/270/commits/b81f61f786960453af2449a016876285e45a2426)

.github/workflows/main.yml
ssolson: What coveralls does is consider lines of code ran vs lines of code tested and aggregating that into a single number. 

When we do not run hindcast the lines do not count against coverage (in our case we will catch this in commits to master). 

Coverage is always a tenuous metric at best.

PR #272, opened 2023-11-02
Remove ci

ssolson:
I believe the ci folder is deprecated now that we use GitHub actions and is an artifact from the original MHKiT testing.


ssolson:
Hey @akeeste I noticed some files that needed to be removed and made this quick PR to address those issues. Could you take a look?

akeeste (approved):
@ssolson Looks good to me! This will clean-up the top-level of the repo nicely

PR #279, opened 2023-11-27
Convert loads module to xarray

akeeste:
Loads/general
- [x] for functions where pandas is an input, allow xarray as an input
- [x] for functions that take pandas or xarray input, convert the internal functionality to be xarray-based
- [x] for functions that now allow xarray input, add a test 
- [x] for functions that output pandas, add a flag to output xarray

Loads/extreme
- [x] for functions where pandas is an input, allow xarray as an input
- [x] for functions that take pandas or xarray input, convert the internal functionality to be xarray-based
- [x] for functions that now allow xarray input, add a test  
- [x] for functions that output pandas, add a flag to output xarray

### DIFF
--- a/mhkit/loads/extreme.py
+++ b/mhkit/loads/extreme.py
@@ -835,7 +835,8 @@ def mler_wave_amp_normalize(wave_amp, mler, sim, k, to_pandas=True):
     if isinstance(mler,pd.DataFrame):
         mler = mler.to_xarray()
     
-    freq = mler.coords['frequency'].values * 2*np.pi
+    coord_names = list(mler.coords)
+    freq = mler.coords[coord_names[0]].values * 2*np.pi
     dw = (max(freq) - min(freq)) / (len(freq)-1)  # get delta
     
     wave_amp_time = np.zeros((sim['maxIX'], sim['maxIT']))
@@ -917,7 +918,8 @@ def mler_export_time_series(rao, mler, sim, k, to_pandas=True):
     if isinstance(mler,pd.DataFrame):
         mler = mler.to_xarray()
     
-    freq = mler.coords['frequency'].values * 2*np.pi
+    coord_names = list(mler.coords)
+    freq = mler.coords[coord_names[0]].values * 2*np.pi
     dw = (max(freq) - min(freq)) / (len(freq)-1)  # get delta
 
     # calculate the series
--- a/mhkit/loads/extreme.py
+++ b/mhkit/loads/extreme.py
@@ -627,10 +627,8 @@ def mler_coefficients(rao, wave_spectrum, response_desired, time_dimension="", t
         Desired response, units should correspond to a motion RAO or
         units of force for a force RAO.
     time_dimension: string (optional)
-        Name of the xarray dimension corresponding to time. Pandas only defines
-        1 dimension, so for Pandas input the index is assumed to be time.
-        If not supplied for xarray input, time is assumed to be the first 
-        dimension.
+        Name of the xarray dimension corresponding to time. If not supplied, 
+        defaults to the first dimension. Does not affect pandas input.
     to_pandas: bool (optional)
         Flag to output pandas instead of xarray. Default = True.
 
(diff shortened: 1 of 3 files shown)


akeeste:
I'm currently working through a few issues with getting the tests to pass. There is a minor bug in the standard deviation calculation in ``loads.general.bin_statistics``. 

The original code uses pandas input. In ``bin_statistics`` the Pandas input, ``data``, is sliced by both ``signal_name`` and ``binnumber``:

https://github.com/MHKiT-Software/MHKiT-Python/blob/4eeb5989d6fcfd7ce9a434cde79092bc83911bbf/mhkit/loads/general.py#L59-L70

However, issues can sometimes arise. If a signal within a Pandas DataFrame is sliced and only has one value (i.e. for i=22 with the test data), the result is a Pandas Series, not another Pandas DataFrame. The std of a pandas Series returns a float, whereas the std of a pandas DataFrame returns a DataFrame of length 1. This results in data that should have std=0 to error in line 68 and results in nan value on Line 70 instead.

```
df = pd.DataFrame([1., 2.])
s = pd.Series([1., 2.])
print(df.std())
print(s.std())
print(df.std()[0])
print(s.std()[0])
```

Instead of calculating the standard deviation manually, I suggest we use ``scipy.stats.binned_statistic`` with statistic='std', just as we do in line 59 above.

akeeste:
@ssolson This PR is ready for review.

akeeste:
Thanks @ssolson I'll go through the review and make updates early next week.

akeeste:
TODO from discussion with @ssolson today:
- [ ] add an optional argument for the user to specify the time dimension, otherwise assume that time is the 1st dimension
- [ ] add argument validation to check that the relevant dimension is correct (time, frequency, etc)

also
- [ ] confirm that example notebooks using the loads module still correctly

mhkit/loads/extreme.py
```diff
@@ -638,58 +642,66 @@ def mler_coefficients(rao, wave_spectrum, response_desired):
...
+    # If input is pandas, convert to xarray
+    if isinstance(wave_spectrum,pd.DataFrame):
+        wave_spectrum = wave_spectrum.to_xarray()
+    
     # convert from Hz to rad/s
+    freq_hz = wave_spectrum.index.values
     freq = freq_hz * (2*np.pi)
-    # change from Hz to rad/s
-    wave_spectrum = wave_spectrum.iloc[:, 0].values / (2*np.pi)
+    key_name = list(wave_spectrum.keys())[0]
```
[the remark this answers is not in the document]
akeeste: I updated this function to take in Pandas series, dataframe or xarray dataarray or dataset. however if a DataFrame or Dataset is input, they are simplified to a series/dataArray, and so can only contain one variable. This should eliminate the issue of calling out the correct variable name with xarray

akeeste:
@ssolson I addressed your review comments and added the option for a user to specify the time dimension for >1D variables. But I don't know a robust way to verify that the user's specified dimension is for example a timeseries, so I think we should leave that for another update.

The user is also limited to inputting a single variable for some functions, like ``mler_coefficients`` where the user should really only input 1 wave spectrum at a time.

mhkit/loads/extreme.py
```diff
@@ -618,15 +620,21 @@ def mler_coefficients(rao, wave_spectrum, response_desired):
...
-        Wave spectral density [m^2/Hz] indexed by frequency [Hz].
+    wave_spectrum: pandas Series, pandas DataFrame, xarray DataArray, or xarray Dataset
+        Wave spectral density [m^2/Hz] indexed by frequency [Hz]. 
+        DataFrame and Dataset inputs should only have one data variable
     response_desired: int or float
         Desired response, units should correspond to a motion RAO or
         units of force for a force RAO.
+    dimension: string (optional)
+        Name of the xarray dimension corresponding to time.
+        If not supplied, time is assumed to be the first dimension.
```
ssolson, line 631: If this is always time then I think "time_dimension" is a more descriptive name and  increases the clarity for the user to the kwargs intended use without reading the docstring.

I think the docstring should explicitly state that this is only used in xarray (it is certainty implied). So I am thinking along the lines of _~"time_dimesion has no impact on pandas type inputs"

ssolson (commented):
@akeeste I think this looks great. I have two minor changes and then I think this is ready to go. First I think "dimension" should be time "time_dimension" to increase the clarity without the need to read the docstring (I also made a small suggestion that we mention the kwarg is not applied to pandas inputs bc while obvious to us I think it may not be to a new user.). Secondly was simply to make a small change to a type check.

mhkit/loads/extreme.py
akeeste: Got it. That makes sense and is a good idea. Thanks for clarifying; I misunderstand in our last discussion. I'll rename the "dimension" input here and in the other relevant functions

akeeste:
Thanks @ssolson. I'll make those changes then merge this PR

ssolson:
@akeeste the CDIP Threads Server seems to be down right now. The test will not pass until it is back up.
https://thredds.cdip.ucsd.edu/

ssolson:
Threads is back up.

Issue #277, opened 2023-11-08
Trouble reading data subset


hevgyrt:
### Describe the bug:
NOTE: This issue is copied from the [dolfyn repo](https://github.com/lkilcher/dolfyn/issues/122) 
I have the two following issues with regards to using the `nens` argument option in the `dolfyn.read` function.

### To Reproduce:

1. If I provide start values larger than 0, like `dat = dolfyn.read('myfile.ad2cp',nens=[3,150])`, I get

```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[70], line 1
----> 1 dat = dolfyn.read('myfile.ad2cp',nens=[3,150])

File ~/miniconda3/envs/ekok/li/python3.9/site-packages/dolfyn/io/api.py:103, in read(fname, userdata, nens, **kwargs)
     99     func_map = dict(RDI=read_rdi,
    100                     nortek=read_nortek,
    101                     signature=read_signature)
    102     func = func_map[file_type]
--> 103 return func(fname, userdata=userdata, nens=nens, **kwargs)

File ~/miniconda3/envs/ekok/lib/python3.9/site-packages/dolfyn/io/nortek2.py:69, in read_signature(filename, userdata, nens, rebuild_index, debug, **kwargs)
     67 d = rdr.readfile(nens[0], nens[1])
     68 rdr.sci_data(d)
---> 69 out = _reorg(d)
     70 _reduce(out)
     72 # Convert time to dt64 and fill gaps

File ~/miniconda3/envs/ekok/lib/python3.9/site-packages/dolfyn/io/nortek2.py:413, in _reorg(dat)
    410 outdat['long_name'].update(dnow['long_name'])
    411 outdat['standard_name'].update(dnow['standard_name'])
    412 cfg['burst_config' + tag] = lib._headconfig_int2dict(
--> 413     lib._collapse(dnow['config'], exclude=collapse_exclude,
    414                   name='config'))
    415 outdat['coords']['time' + tag] = lib._calc_time(
    416     dnow['year'] + 1900,
    417     dnow['month'],
   (...)
    421     dnow['second'],
    422     dnow['usec100'].astype('uint32') * 100)
    423 tmp = lib._beams_cy_int2dict(
    424     lib._collapse(dnow['beam_config'], exclude=collapse_exclude,
    425                   name='beam_config'), 21)

[...]

PR #280, opened 2023-11-27
Nortek Signature Reader Altimeter Update

jmcvey3:
PR containing fixes for Issue #277 & #284


ssolson:
@jmcvey3 thanks for putting this fix together! Did you see the changes caused a couple errors in the dolfyn tests?

jmcvey3:
@ssolson Yes, I wanted to make sure the issue was solved before going through and fixing the tests. I've updated the tests now that I've gotten the affirmative that the issue was resolved.

I also threw in a fix for a bug where netcdf compression wasn't working properly. The prior code would only work with netCDF4 < 1.6.0, which isn't compatible with python 3.11.

mhkit/dolfyn/io/nortek_defs.py
```diff
@@ -340,3 +340,144 @@ def sci_func(self, data):
...
+                         standard_name='time',
+                         ),
+    'batt_alt': _VarAtts(dims=[],
+                         dtype=np.uint16,
+                         group='data_vars',
+                         default_val=nan,
+                         factor=0.1,
+                         units='V',
+                         long_name='Battery Voltage',
+                         ),
```
ssolson, line 360: this one doesn't have a `standard_name`

mhkit/dolfyn/io/nortek_defs.py
```diff
@@ -340,3 +340,144 @@ def sci_func(self, data):
...
+
+waves_hdrdata = {
+    'time_alt': _VarAtts(dims=[],
+                         dtype=np.float64,
+                         group='coords',
+                         default_val=nan,
+                         units='seconds since 1970-01-01 00:00:00 UTC',
+                         long_name='Time',
+                         standard_name='time',
+                         ),
```
ssolson, line 352: should this have a `factor` key?

mhkit/dolfyn/io/nortek_defs.py
```diff
@@ -340,3 +340,144 @@ def sci_func(self, data):
...
+                          factor=0.001,
+                          units='m',
+                          long_name='AST distance2 on vertical beam',
+                          standard_name='altimeter_range',
+                          ),
+    'AnaIn1_alt': _VarAtts(dims=[],
+                           dtype=np.float32,
+                           group='sys',
+                           default_val=nan,
+                           units='n/a',
```
ssolson, line 458: Should this one have no name keys?

mhkit/dolfyn/io/nortek_defs.py
```diff
@@ -340,3 +340,144 @@ def sci_func(self, data):
...
+                           ),
+    'vel_alt': _VarAtts(dims=[4, 'n'],
+                        dtype=np.float32,
+                        group='data_vars',
+                        default_val=nan,
+                        factor=0.001,
+                        units='m s-1',
+                        long_name='Water Velocity',
+                        ),
```
ssolson, line 467: standard_name?

mhkit/dolfyn/io/nortek_defs.py
```diff
@@ -340,3 +340,144 @@ def sci_func(self, data):
...
+                        long_name='Acoustic Signal Amplitude',
+                        standard_name='signal_intensity_from_multibeam_acoustic_doppler_velocity_sensor_in_sea_water',
+                        ),
+    'quality_alt': _VarAtts(dims=[],
+                            dtype=np.float32,
+                            group='data_vars',
+                            default_val=nan,       
+                            units='1',
+                            long_name='Altimeter Quality Indicator',
+                            ),
```
ssolson, line 482: standard_name?

ssolson (commented):
Thank @jmcvey3 this is looking pretty much done. I only really spotted a couple of questions in the new dictionaries you added. 

Could you briefly document here what you updated in all the bin files?

jmcvey3:
> Thank @jmcvey3 this is looking pretty much done. I only really spotted a couple of questions in the new dictionaries you added.
> 
> Could you briefly document here what you updated in all the bin files?

Given new test data, I found that the altimeter data wasn't being labeled properly versus what Nortek software was outputting. These changes primarily handle those metadata updates, and because the metadata changes, the test files also change.

mhkit/dolfyn/io/nortek_defs.py
jmcvey3: Variables don't get standard names if they aren't listed in this table: https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html

mhkit/dolfyn/io/nortek_defs.py
jmcvey3: No, the factor key here is for any variables that are stored in binary in something other than their native units. Time is stored in a binary format that converts directly to "seconds since 1970"

mhkit/dolfyn/io/nortek_defs.py
jmcvey3: This variable isn't used, so I don't add them. This is the raw ADC output, I believe

mhkit/dolfyn/io/nortek_defs.py
jmcvey3: Same comment as above

mhkit/dolfyn/io/nortek_defs.py
jmcvey3: Same comment as above

PR #282, opened 2023-12-18
Convert power module to xarray

akeeste:
power.quality
- [x] for functions where pandas is an input, allow xarray as an input
- [x] for functions that take pandas or xarray input, convert the internal functionality to be xarray-based
- [x] for functions that now allow xarray input, add a test 
- [x] for functions that output pandas, add a flag to output xarray

power.characteristics
- [x] for functions where pandas is an input, allow xarray as an input
- [x] for functions that take pandas or xarray input, convert the internal functionality to be xarray-based
- [x] for functions that now allow xarray input, add a test  
- [x] for functions that output pandas, add a flag to output xarray

### DIFF
--- a/mhkit/power/characteristics.py
+++ b/mhkit/power/characteristics.py
@@ -8,7 +8,6 @@ def instantaneous_frequency(um, to_pandas=True):
 
     """
     Calculates instantaneous frequency of measured voltage
-    
      
     Parameters
     -----------
@@ -25,7 +24,8 @@ def instantaneous_frequency(um, to_pandas=True):
         with signal name columns
     """  
     if not isinstance(um, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
-        raise TypeError(f'um must be of type pd.Series or pd.DataFrame. Got: {type(um)}')
+        raise TypeError('um must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(um)}')
    
     if isinstance(um.index[0], datetime.datetime):
         t = (um.index - datetime.datetime(1970,1,1)).total_seconds()
@@ -74,9 +74,11 @@ def dc_power(voltage, current, to_pandas=True):
         DC power [W] from each channel and gross power indexed by time
     """
     if not isinstance(voltage, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
-        raise TypeError(f'voltage must be of type pd.Series or pd.DataFrame. Got: {type(voltage)}')
+        raise TypeError('voltage must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(voltage)}')
     if not isinstance(current, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
-        raise TypeError(f'current must be of type pd.Series or pd.DataFrame. Got: {type(current)}')
+        raise TypeError('current must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(current)}')
     if not voltage.shape == current.shape:
         raise ValueError('current and volatge must have the same shape')
     
@@ -122,9 +124,11 @@ def ac_power_three_phase(voltage, current, power_factor, line_to_line=False, to_
         Magnitude of active AC power [W] indexed by time with Power column 
     """
     if not isinstance(voltage, (pd.DataFrame, xr.Dataset)):
-        raise TypeError(f'voltage must be of type pd.DataFrame. Got: {type(voltage)}')
+        raise TypeError('current must be of type pd.DataFrame ' + 
+                        f'or xr.Dataset. Got{type(current)}')
     if not isinstance(current, (pd.DataFrame, xr.Dataset)):
-        raise TypeError(f'current must be of type pd.DataFrame. Got: {type(current)}')
+        raise TypeError('current must be of type pd.DataFrame ' + 
+                        f'or xr.Dataset. Got{type(current)}')
     if not len(voltage.columns) == 3:
         raise ValueError('voltage must have three columns')
     if not len(current.columns) == 3:
--- a/mhkit/power/characteristics.py
+++ b/mhkit/power/characteristics.py
@@ -4,7 +4,7 @@ import numpy as np
 from scipy.signal import hilbert
 import datetime
 
-def instantaneous_frequency(um, to_pandas=True):
+def instantaneous_frequency(um, dimension="", to_pandas=True):
 
     """
     Calculates instantaneous frequency of measured voltage
@@ -13,6 +13,10 @@ def instantaneous_frequency(um, to_pandas=True):
     -----------
     um: pandas Series, pandas DataFrame, xarray DataArray, or xarray Dataset
         Measured voltage (V) indexed by time 
+
+    dimension: string (optional)
+        Name of the xarray dimension corresponding to time.
+        If not supplied, time is assumed to be the first dimension.
         
     to_pandas: bool (Optional)
         Flag to save output to pandas instead of xarray. Default = True.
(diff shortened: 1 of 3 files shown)


mhkit/power/characteristics.py
```diff
@@ -1,121 +1,185 @@
...
-    if not isinstance(current, pd.DataFrame):
-        raise TypeError(f'current must be of type pd.DataFrame. Got: {type(current)}')
-    if not len(voltage.columns) == 3:
+    if not isinstance(voltage, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('voltage must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(voltage)}')
+    if not isinstance(current, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('current must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(current)}')
+
```
ssolson, line 148: We should add type checks to `line_to_line` and `to_pandas`

mhkit/power/characteristics.py
```diff
@@ -1,121 +1,185 @@
...
-    if not current.shape == voltage.shape:
+    if not current.sizes == voltage.sizes:
         raise ValueError('current and voltage must be of the same size')
-    
 
-    abs_current = np.abs(current.values)
-    abs_voltage = np.abs(voltage.values)
+    power = dc_power(voltage, current, to_pandas=False)['Gross']
+    power.name = 'Power'
+    power = power.to_dataset() # force xr.DataArray to be consistently in xr.Dataset format
```
ssolson, line 163: This forcing seems like a intermediate step needed prior to getting dc_power to return an xarray.

Is this still needed when we have `to_pandas=False`?

mhkit/power/characteristics.py
```diff
@@ -1,121 +1,185 @@
...
-    if not isinstance(um, (pd.Series, pd.DataFrame)):
-        raise TypeError(f'um must be of type pd.Series or pd.DataFrame. Got: {type(um)}')
-   
-    if isinstance(um.index[0], datetime.datetime):
-        t = (um.index - datetime.datetime(1970,1,1)).total_seconds()
-    else:
-        t = um.index
+    if not isinstance(um, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('um must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(um)}')
```
ssolson, line 31: I think we should add type checking to `time_dimension` and `to_pandas`

mhkit/power/characteristics.py
```diff
@@ -1,121 +1,185 @@
...
+
+    if to_pandas:
+        P = P.to_pandas()
+
     return P
+
+def _convert_to_dataset(data, name='data'):
+    # Takes data that could be pd.DataFrame, pd.Series, xr.DataArray, or 
+    # xr.Dataset and converts it to xr.Dataset
+    if isinstance(data, (pd.DataFrame, pd.Series)):
```
ssolson, line 177: Let's give this a formal docstring and type checking for production.

    """
    Converts the given data to an xarray.Dataset.

    This function is designed to handle inputs that can be either a pandas DataFrame, a pandas Series,
    an xarray DataArray, or an xarray Dataset. It ensures that the output is consistently an xarray.Dataset.

    Parameters
    ----------
    data: pandas DataFrame, pandas Series, xarray DataArray, or xarray Dataset
        The data to be converted. 

    name: str (Optional)
        The name to assign to the data variable in case the input is an xarray DataArray without a name.
        Default value is 'data'.

    Returns
    -------
    xarray.Dataset
        The input data converted to an xarray.Dataset. If the input is already an xarray.Dataset,
        it is returned as is.

    Examples
    --------
    >>> df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]})
    >>> ds = _convert_to_dataset(df)
    >>> type(ds)
    <class 'xarray.core.dataset.Dataset'>

    >>> series = pd.Series([1, 2, 3], name='C')
    >>> ds = _convert_to_dataset(series)
    >>> type(ds)
    <class 'xarray.core.dataset.Dataset'>

    >>> data_array = xr.DataArray([1, 2, 3])
    >>> ds = _convert_to_dataset(data_array, name='D')
    >>> type(ds)
    <class 'xarray.core.dataset.Dataset'>
    """
    if not isinstance(data, (pd.DataFrame, pd.Series, xr.DataArray, xr.Dataset)):
        raise TypeError("Input data must be of type pandas.DataFrame, pandas.Series, "
                    "xarray.DataArray, or xarray.Dataset")
                        
    if not isinstance(name, str):
        raise TypeError("The 'name' parameter must be a string")

mhkit/power/quality.py
```diff
@@ -20,197 +20,233 @@ def harmonics(x, freq, grid_freq):
...
+    if not isinstance(x, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('x must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(x)}')
 
     if not isinstance(freq, (float, int)):
-        raise ValueError('freq must be of type float or integer')
+        raise TypeError('freq must be of type float or integer')
 
     if grid_freq not in [50, 60]:
         raise ValueError('grid_freq must be either 50 or 60')
```
ssolson, line 41: Let's do a `to_pandas` type check

mhkit/power/quality.py
```diff
@@ -20,197 +20,233 @@ def harmonics(x, freq, grid_freq):
...
     """
-    if not isinstance(harmonics, (pd.Series, pd.DataFrame)):
-        raise ValueError('harmonics must be of type pd.DataFrame or pd.Series')
-
+    if not isinstance(harmonics, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('harmonics must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(harmonics)}')
+    
     if grid_freq not in [50, 60]:
         raise ValueError('grid_freq must be either 50 or 60')
```
ssolson, line 105: Let's add `frequency_dimension` and  `to_pandas` type checks

mhkit/power/quality.py
```diff
@@ -20,197 +20,233 @@ def harmonics(x, freq, grid_freq):
...
-
-    if not isinstance(rated_current, float):
-        raise ValueError('rated_current must be a float')
-
-    harmonics_sq = harmonics_subgroup.iloc[2:50]**2
-
+    if not isinstance(harmonics_subgroup, (pd.Series, pd.DataFrame, xr.DataArray, xr.Dataset)):
+        raise TypeError('harmonics_subgroup must be of type pd.Series, pd.DataFrame, ' + 
+                        f'xr.DataArray, or xr.Dataset. Got{type(harmonics_subgroup)}')
+
```
ssolson, line 165: type checks for new inputs

mhkit/tests/power/test_power.py
```diff
@@ -42,66 +43,130 @@ def setUpClass(self):
...
 
-    def test_TCHD_sine_wave(self):
-        current = pd.Series(self.signal, index=self.samples)
+    def test_TCHD_sine_wave_pandas(self):
         harmonics = pd.DataFrame(self.harmonics_vals,
-                                 index=self.harmonics_int)
+                                  index=self.harmonics_int)
         hsg = power.quality.harmonic_subgroups(harmonics, self.frequency)
+        TCHD = power.quality.total_harmonic_current_distortion(
+            hsg)  # had to just put a random rated current in here
```
ssolson, line 83: Let's remove all instances of this comment while we are here and document somewhere in set up class or something is its really needed:
> `# had to just put a random rated current in here`

mhkit/tests/power/test_power.py
```diff
@@ -42,66 +43,130 @@ def setUpClass(self):
...
+
+    def test_dc_power_pandas(self):
+        current = pd.DataFrame(self.current_data, columns=['A1','A2','A3'])
+        voltage = pd.DataFrame(self.voltage_data, columns=['V1','V2','V3'])
         P = power.characteristics.dc_power(voltage, current)
-        self.assertEqual(P.sum()['Gross'],
-                         (voltage.values * current.values).sum())
+        self.assertEqual(P.sum()['Gross'], 584)
+        P = power.characteristics.dc_power(voltage['V1'], current['A1'])
+        self.assertEqual(P.sum()['Gross'], 70)
```
ssolson, line 135: It looks like you changed this from recalculating the values to hardcoding them. 

Are we getting the same values here as before or could you just comment on this change in general?

ssolson (commented):
@akeeste thank you for putting this together and great job. In tis review I basically want to 

1. Add type checks to the paramters this PR adds
2. Minor clean up from previous authors
3. Couple of minor questions

I still need to run through the example notebook tomorrow

mhkit/tests/power/test_power.py
akeeste: Good call. Also i actually just removed the rated current input all together, because it was not being used in anyway. I'm wondering if we're missing a scaling or something for THCD that would use the rated current

mhkit/power/characteristics.py
akeeste: Agreed, done. Might have to move this to a more general location later on

mhkit/power/characteristics.py
akeeste: When the Gross power variable is extracted from the results of dc_power, the xr.Dataset automatically becomes an xr.DataArray since there's just one variable. I originally added this conversion to a xr.Dataset to prevent that type change.

There shouldn't be any issues with returning a single variable array like pd.Series or xr.DataArray, but the previous function and most of the current ones I'm working with return pd.DataFrame and xr.Dataset

mhkit/tests/power/test_power.py
akeeste: Reviewing the definition of THCD (not with the marine energy IEC standard), I don't see a rated current used. From what I can tell, THCD is the ratio (in percent) between the energy of the harmonic subgroups and the energy of the dominant frequency and doesn't require a rated current.

mhkit/tests/power/test_power.py
akeeste: These are exactly the same values that the test should give, just hard coded in. It's been a few weeks, I think I was having issues duplicating the recalculation method easily across pandas and xarray because of their different formats. Looking at it again, I think a better solution is to recalculate the power values, but with self.current_data and self.voltage_data. That would eliminate any differences between xarray/pandas and be easily repeatable across both tests

akeeste:
@ssolson Thanks for the review. I went through your comments and resolved them, which were primarily on adding parameter validation and doc strings where missing. I reverted the test to recalculate the answer as described above, and double-checked that the power notebook runs. I think the only outstanding item is to ensure that the THCD calculation is not missing anything.

Issue #287, opened 2024-01-31
(Dolfyn) Non-identical beam error on multiple Nortek Signature 500 .ad2cp files?


willcoxe:
Hello, Thanks for creating the package!

I just received several Nortek Signature 500 ADCP files to process and each of them gives the same error (details below). I have updated to the most recent version and tried to call the command in several ways. I have tested the example signature files in the examples directory and they can be read without error. I am hoping you might have some advice on troubleshooting this issue. 

Commands  I have tried (I have replaced file and directory names):
```
>>> ds = api.read("FILE.ad2cp")
>>> ds = dolfyn.read("FILE.ad2cp")
>>> ds = dolfyn.io.nortek2.read_signature("FILE.ad2cp")
>>> dolfyn.io.nortek2.read_signature("FILE.ad2cp", rebuild_index=True)
```

While running it skips a lot of pings e.g.

```
warnings.warn("Skipped ping (ID: {}) in file {} at ensemble {}."
venvdir/site-packages/mhkit/dolfyn/io/nortek2_lib.py:176: UserWarning: Skipped ping (ID: 31) in file FILE.ad2cp at ensemble 64257.
```

The message it errors out with is always the same:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "venvdir/site-packages/mhkit/dolfyn/io/api.py", line 105, in read
    return func(fname, userdata=userdata, nens=nens, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venvdir/site-packages/mhkit/dolfyn/io/nortek2.py", line 68, in read_signature
    rdr = _Ad2cpReader(filename, rebuild_index=rebuild_index, debug=debug)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venvdir/site-packages/mhkit/dolfyn/io/nortek2.py", line 136, in __init__
    self._config = lib._calc_config(self._index)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venvdir/site-packages/mhkit/dolfyn/io/nortek2_lib.py", line 493, in _calc_config
    raise Exception("beams_cy are not identical for id: 0x{:X}."
Exception: beams_cy are not identical for id: 0x17.
```

[...]

PR #281, opened 2023-11-29
Black Formatting

ssolson:
This PR adds starts to address formatting issues in MHKiT by adding black:
- Black is an auto-formatter, not a Python linter or interpreter. 
- https://github.com/psf/black

This PR adds black because it will automatically format our files allowing our users to focus on coding and letting black fix the formatting issues automatically.

To adhere to black formatting developers can install and run black from the terminal manually, they can install black as an extension in their IDE such as VS code, or they can enable enable the pre-commit hook.

What this PR changes:
- Ran  `black .`  formatting on the repo
  - After installing black I auto reformatted the repository using the above command
- add an GitHub actions check that code adheres to black
- Added an optional pre commit that will auto format files to adhere to black on commits should a developer choose to use it. 
- Added test coverage so that tests would pass

PR #289, opened 2024-02-09
Add capability to read ADCP dual profile configurations

jmcvey3:
Fix for #287, which is data from a dual profiling ADCP taking a large combination of burst and average measurements.


ssolson:
@jmcvey3 as of #281 all MHKiT code must adhere to black formatting.

You simply need to install the package and run `black .`

You can also integrate it into your IDE such as VSCode or create a pre-commit hook as detailed in the MHKiT README.

jmcvey3:
Will fix tests when I know everything is functional

jmcvey3:
@ssolson Can you see if you can format the two files this test says are failing? Running `black .` makes no further changes beyond commit 7fa66b0

ssolson:
> @ssolson Can you see if you can format the two files this test says are failing? Running `black .` makes no further changes beyond commit [7fa66b0](https://github.com/MHKiT-Software/MHKiT-Python/commit/7fa66b027d2f61723613ce05fc9250e5acf230d3)

The error says to install black[jupyter]
I'm not sure why since these are not ipynb but can you try that?

image

jmcvey3:
> > @ssolson Can you see if you can format the two files this test says are failing? Running `black .` makes no further changes beyond commit [7fa66b0](https://github.com/MHKiT-Software/MHKiT-Python/commit/7fa66b027d2f61723613ce05fc9250e5acf230d3)
> 
> The error says to install black[jupyter] I'm not sure why since these are not ipynb but can you try that?
> 
> ...ohh wait that is just a warning of sorts... I will try tomorrow.
> 
> One other thing to check is that your black formatter is the latest version.
> 
> image

Hmm, just uninstalled the pip version and installed it from conda. Still no luck.

ssolson:
@jmcvey3 I upgraded my black using pip
`pip install black --upgrade`

I was able to then update using `black .`

I would try the pip version since releases are faster vs conda and out tests use a pip install.

If you don't care about that I pushed to `ssolson dual_profile` which you can pull from to just knock this task out.

image

jmcvey3:
@willcoxe Is it alright with you if we use the 1 Mb file for a software test? I will try to trim it down to the minimum file size necessary to check all the code I updated.

willcoxe:
I don't own the file so I can't make that call I'm afraid. It's eventually intended to be published with a doi by the uni/research group who own it.

mhkit/dolfyn/io/nortek2_lib.py
```diff
@@ -242,16 +261,20 @@ def _check_index(idx, infile, fix_hw_ens=False):
...
+                FLAG = True
+                hwe[id == idx["ID"]] += d
+                ens[id == idx["ID"]] += d
+
+        if np.any(np.diff(ens) > 1) and FLAG:
+            idx["ens"] = np.unwrap(hwe.astype(np.int64), period=period) - hwe[0]
 
-    if np.any(np.diff(ens) > 1) and FLAG:
-        idx["ens"] = np.unwrap(hwe.astype(np.int64), period=period) - hwe[0]
+    return dp
```
ssolson, line 277: Its not clear to me what is going on in this function. Why are we doing all these calculations to return a boolean dp that is not modified? Is this modifying the idx data structures in place so that we do not need to return those values (if so shouldn't we be more explicit about what this function is doing)?

mhkit/dolfyn/io/nortek2.py
```diff
@@ -34,6 +40,8 @@ def read_signature(
       Default = False
     debug : bool
       Logs debugger ouput if true. Default = False
+    dual_profile : bool (default: True)
+      Set to true if instrument is running multiple profiles
```
ssolson, line 44: The default is set to `False`

ssolson (commented):
James thanks for getting htis in. I have just a couple of comments on my read through of this PR.

requirements.txt
```diff
@@ -11,7 +11,7 @@ NREL-rex>=0.2.63
 six>=1.13.0
 h5py>=3.6.0
 h5pyd>=0.7.0
-netCDF4>=1.5.8
+netCDF4
```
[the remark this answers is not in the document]
jmcvey3: Good catch, this may have been git's auto-merge.

mhkit/dolfyn/io/nortek2_lib.py
jmcvey3: I can try removing this bit of code and see if that breaks tests, because I find a little more redundant code each time I need to make changes to this reader. 
Unfortunately a lot of the RDI reader modifies data structures in place; if I had to guess the main reason this was done is to conserve memory for very large files (>1 Gb).

jmcvey3: A workaround for the in-place modification may be to add these data structures to the class (i.e. `self.idx`)

ssolson: Thanks for clarifying. 

I think we can both agree that modifying in place and not being explicit about it is confusing and not best practice. 

Long term I would like us to make a plan which cleans this up but short term let me know what you think is best which can be to leave as is and explicitly address in the future.

jmcvey3: Sounds good. In the interest of merging this PR, and knowing there is a lot of undocumented in-place modification in these readers, let's address it in a single PR in the future.

jmcvey3: Turns out that blip of code is unused.

Issue #286, opened 2024-01-30
Dolfyn Reading Online Singature Deploment Measurments 


browniea:
Using the Nortek "Signature Deployment" software the online measurement output (.AD2CP files)  have errors being read by `dolfyn.read`

note: Using the same process but a deployment vs online measurement has worked with no issues so far.

jmcvey3:
Was able to fix one of the files so far in https://github.com/jmcvey3/MHKiT-Python/tree/online, the one ending in "0943". The file ending in "4835" also already runs successfully.

jmcvey3:
@browniea Found a way to skip past the header data. See if that works for you

PR #290, opened 2024-02-19
Skip junk headers in AD2CP files

jmcvey3:
Adds logic to skip the messy header data that can accumulate during measurements collected via the Nortek software on a PC/Mac. Fix for #286


ssolson:
@browniea could you give this PR a review and let us know what you find?

browniea (approved):
One last data file gave me a memory error that I sent over email. Other than that It worked for every other file I threw at it.

Issue #291, opened 2024-02-22
DOLfYN range issue 


browniea:
When importing data from the Teledyne River Pro ADCP the range of depth bins doesn't load the complete list. the first 'range' screenshot is what DOLfYN imports (before being properly offset for deployment depth). The second screenshot is the correct bin depths in rows of time steps.  

![unnamed](https://github.com/MHKiT-Software/MHKiT-Python/assets/63879572/c4f1f610-fad8-4a38-a548-1a399856e60f)
![unnamed](https://github.com/MHKiT-Software/MHKiT-Python/assets/63879572/c22037d6-bb8b-46fd-b3ec-4e653e9751f9)

jmcvey3:
One more fix for this in #294

PR #292, opened 2024-02-22
Handle RDI's variable profile shenanigans

jmcvey3:
Introduces fixes for RDI files that contain changing "number of cells" and variable "cell sizes". The latter are bin-averaged into the largest cell size. Issue #291


jmcvey3:
@browniea When you get a chance, see if this works for you. The range may still offset based on whatever the "bin1_dist_m" attribute, but that should be easily adjusted.

browniea:
@jmcvey3  Looks like there is a conflict with the latest merge [#286](https://github.com/MHKiT-Software/MHKiT-Python/issues/286)

jmcvey3:
> @jmcvey3 Looks like there is a conflict with the latest merge [#286](https://github.com/MHKiT-Software/MHKiT-Python/issues/286)

Yeah that's annoying

browniea:
The variable bin size is still being missed but the number of bins is right! 
image

jmcvey3:
> The variable bin size is still being missed but the number of bins is right! image

Sweet. The data from the 0.12 m bins is stored in the variables with an "_sl" tag, so I'm sure it wouldn't be difficult to write a function to combine these two. However, I don't think that function would belong in the reader, since the goal of the reader is to read what's in the datafile and do no more processing than necessary.

browniea:
That makes sense and I agree.  I can include that in my next example notebook for UMERC stuff.

browniea (approved):

jmcvey3:
> That makes sense and I agree. I can include that in my next example notebook for UMERC stuff.

Great, yes I think that's a good idea, especially since TRDI is retiring the Workhorse line and pushing these newer instruments.

PR #293, opened 2024-03-05
Turbulence intensity and instrument noise

jmcvey3:
Adds a function so that instrument noise can be subtracted from turbulence intensity estimation. This function also uses RMS to calculate TI, and it's output differs by about 1% from the pre-existing "I" property, which uses standard deviation.

Fixed a bug where a given noise value was not being subtracted from the power spectra, and noise was added as an input to dissipation rate calculations.


jmcvey3:
@ssolson This covers what I mentioned earlier

mhkit/dolfyn/velocity.py
```diff
@@ -913,10 +968,9 @@ def power_spectral_density(
         window : string or array
           Specify the window function.
           Options: 1, None, 'hann', 'hamm'
-        noise : float or array-like
-          A vector of the noise levels of the velocity data with
-          the same first dimension as the velocity vector.
-          Default = 0.
+        noise : numeric or array
+          Instrument noise level in same units as velocity.
+          Default: 0 (ADCP) or [0, 0, 0] (ADV).
```
ssolson, line 973: The actual default is None

jmcvey3: Hmmm, yes changing that to 0 is better for code transparency

PR #307, opened 2024-04-24
MHKiT v0.8.0

ssolson:
# MHKiT v0.8.0
We're excited to announce the release of MHKiT v0.8.0, which brings a host of new features, enhancements, and bug fixes across various modules, ensuring compatibility with Python 3.10 and 3.11, and introducing full xarray support for more flexible data handling. Significant updates in the Wave and DOLfYN modules improve functionality and extend capabilities.

## Python 3.10 & 3.11 Support
MHKiT now supports python 3.10 and 3.11. Support for 3.12 will follow in the next minor update.
- https://github.com/MHKiT-Software/MHKiT-Python/pull/240

### Enhancements:
**Automatic Threshold Calculation for Peaks-Over-Threshold**: We've introduced a new feature that automatically calculates the "best" threshold for identifying significant wave events. This method, originally developed by Neary, V. S., et al. in their 2020 study, has now been translated from Matlab to Python, enhancing our existing peaks-over-threshold functionality.

**Wave Heights Analysis**: A new function, `wave_heights`, has been added to extract the heights of individual waves from a time series. This function uses zero up-crossing analysis to accurately measure wave heights, improving upon our previous methods which only provided the maximum value between up-crossings.

**Enhanced Zero Crossing Analysis**: Building on the above, the zero crossing code previously embedded in `global_peaks` has been isolated into a helper function. This modular approach not only refines the codebase but also supports new functionalities such as calculating wave heights, zero crossing periods, and identifying crests.

### Bug Fixes:
**Contour Sampling Error in Wave Contours**: A bug identified in `mhkit.wave.contours.samples_contour` has been resolved. The issue occurred when period samples defined using the maximum period resulted in values outside the interpolation range of the contour data. This was corrected by ensuring that all sampling points are within the interpolation range and adjusting the contour data selection process accordingly.

- https://github.com/MHKiT-Software/MHKiT-Python/pull/268 
- https://github.com/MHKiT-Software/MHKiT-Python/pull/252 
- https://github.com/MHKiT-Software/MHKiT-Python/pull/278

## Xarray Support
MHKiT functions now fully support the use of xarray for passing and returning data.

- https://github.com/MHKiT-Software/MHKiT-Python/pull/279 
- https://github.com/MHKiT-Software/MHKiT-Python/pull/282
- https://github.com/MHKiT-Software/MHKiT-Python/pull/285
- https://github.com/MHKiT-Software/MHKiT-Python/pull/302
[...]

PR #311, opened 2024-04-25
Fix hanging Hindcast tests

ssolson:
This PR addresses 2 issues with the hindcast tests:
1. On PRs tests were only being run on changes to hindcast test files.
    - Fixed to run on changes to hindcast test and module files.
2. The `prepare cache` job was failing due to 6 hour job limit.
    - Fixed by breaking out each hindcast cache test into its own job.

Additionally this PR updates action packages to the latest version where possible.


ssolson:
Okay we are for sure hitting a 6 hour limit of the wind tests.

image

Running the tests locally it takes 4 hours and 22 minutes. Which is pretty insane amount of time. I think if I break these out into their own jobs I could get 6 hours per job, 

image

ssolson:
Okay great the hindcast jobs now run separate which should fix the 6 hour issue. The new job flow looks as follows:
image

.github/workflows/main.yml
```diff
@@ -30,15 +30,15 @@ jobs:
...
 
       - name: Check for changes in wave/io/hindcast
         id: changes
-        uses: dorny/paths-filter@v2
+        uses: dorny/paths-filter@v3
         with:
           filters: |
             wave_io_hindcast:
-              - 'mhkit/tests/wave/io/hindcast/**'
+              - 'mhkit/wave/io/hindcast/**'
```
akeeste, line 41: we should probably check for changes in both places: hindcast tests and the hindcast module itself

ssolson: Thanks I have been too lazy to investigate if I could just add the test line to the filters.  Looks like I can.

I'll push the commit up after the current actions pass or fail.

ssolson:
Okay great new test structure is working. Just need to check these final adjustments to make sure the check changes on multiple file locations and bumping up the versions on the actions works.

ssolson:
@akeeste with the tests now working could you give this a final review/ approval (assuming all tests pass in ~4 hours)? 

My goal is to get this into develop today/ this weekend so I can then have these tests pass on #307 over the weekend.

PR #313, opened 2024-04-26
Fix: Handle unnamed index/dimension in `wave.resource.surface_elevation`

simmsa:
Fix: If unnamed, set spectrum `dim_0` name to "Frequency"

The `surface_elevation` function typically takes in a spectrum generated
by an mhkit function that sets the index/dimension name to "Frequency".
This fix handles an atypical case where the index/first dimension is
unnamed, which causes downstream issues where we are expecting a
dimension named "Frequency".

Per the [xarray DataArray documentation](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html)
an unnamed dimension starts at `dim_0`. If we find this case we
change the `dim_0` name to "Frequency".


ssolson:
Hey @jmcvey3 apologies for messing this PR up. This should be easier going forward but there were some inconsistencies in how PR merged strategies were applied in the previous release by the final PR merge author which made this messier. I perhaps should have just accepted the mess but instead decided to squash into master. My preferred strategy going forward is to squash PRs to develop and merge by ort into master.

To ensure master and develop are even I did a force rebase on origin develop.

That said can you either rebase or resubmit this PR? 
The official approach is to:
```
git checkout develop
git fetch origin
git reset --hard origin/develop

git checkout feature-branch
git rebase develop
git push user-fork feature-branch --force
```

That said rebasing may be extremely painful and it might be easier to just start from the new develop and remake the changes. Apologies for the trouble on this.

PR #317, opened 2024-04-29
Fix for failing conda tests

ssolson:


ssolson:
@akeeste I am planning to merge this as soon as it passes so we can get started on the 4+ hour testing run. 

If you see any issue with these changes please lmk here if this is still open on on #307 if I have merged this and I am happy to address in a follow on PR.

PR #318, opened 2024-04-30
Hindcast Test: Conda python version

ssolson:
- Uses python version correctly
  - fixes an incorrect python version setting from #311 
- Installs MHKiT dependencies in conda build 
  - (removes `--no-deps` flag)
- Use coverage actions instead of coverage CLI 
  - some jobs were giving issues with the coverage CLI, coverage action fixed issues and is a more modular approach to coverage.
- Uses an updated coverage version with lcov support
  - to use the coverage action we need an lcov reprot. The default anaconda channel uses a very old version of coverage which I needed to set to the current version of coverage.
- Adds an `environment.yaml` file for conda environment install 
  - this is cleaner and cleanly  separates MHKiT environment packages from tersting packages
  - This is also useful conda users who want to create an MHKiT environment 
- Special case for MacOS-latest (macos 14)  and Py 3.8
  - Implemented solution: use homebrew to install netcdf and hdf5
  - Other options were use macos13 or drop python 3.8 support
- CI test job `set-os` logic modified to now run all OS on push to `develop`
  - previously ran only `ubuntu-latest` on push and PR, 
  - this change keeps with the intention of fastest PR test runs but now on pushes we can check all OS and catch the issues this PR has been fixing sooner.


ssolson:
@akeeste I am finalizing some clean up on this PR and if you are around this is ready for a review. 

I will change the set-os job back to only run ubuntu-latest before merging.

PR #326, opened 2024-05-08
Improve Reynolds stress ADCP estimation notebook discussion V2

jmcvey3:
Editing language based on more insight I've gained from reading literature on ADCP Reynolds stress and TKE measurements. The main edits here are to the adcp_example notebook and the removal of the "total_tke" function. I also updated the ADV notebook to cover all the functions

The data you get from ADCPs requires two critical assumptions. 1. There is next to no vertical motion from the time a parcel of water passes through one beam before it passes through another ("assumption of homogeneity" in the lit). 2. One can't make inferences of turbulent scales smaller than the distance between each of the beams.

ADCPs seem to struggle to measure the u'u' and v'v' TKE components because assumption 1 isn't always true. The second assumption renders the total TKE less useful because the information contained in smaller turbulent scales, which might be critical, is lost.

I wrote the function maybe 8 months ago because TKE is one of the parameters our modeling team uses for validating tidal models. I ultimately want to remove it because my experiences with it so far have been quite off from ADV measurements and model output. The debate is still ongoing on how good ADCP estimations are for TKE/Reynolds stresses, but it likely depends on the measurement site.

ADVs don't have either of these problems because they're making a point measurement, and are still considered the de facto "ground truth" when analyzing turbulence data. (Shear probes also fall under this category.) The calc_tke function under the ADVBinner outputs each of the TKE components; it's trivial to add the components together to get total TKE.


ssolson (approved):
This PR replaces #316

jmcvey3:
Some bugfixes for the dudz, dvdz, dwdz functions incoming...

jmcvey3:
Waiting for the test fixes in the develop branch get completed before merging

ssolson:
@jmcvey3 with your last push you are fixing numpy 2.0 issues. 

Could you help me out by addressing the remaining failing tests in PR #333?

The Dolfyn tests seem to be hit the hardest by this upgrade. Once we get #333 settled we can move the rest of these PRs forward.

PR #330, opened 2024-06-14
ipython notebook test

### LINKED ISSUE #329: Example Testing
Add Action to test example notebooks.

ssolson:
This PR adds a Github action to test the example notebooks as part of or CD pipeline. Additionally a timeout is added to which notebooks will fail if they exceed the given time.

To accomplish this:

1. main.yml - add a job to dynamically gather all *.ipynb and set a timeouts for the notebooks
  - `generate_notebook_matrix.py` is a new script called by the above job
3. main.yml - add a job to run each notebook (skip hindcast if not run, fail on timeout)
4. Update *.ipynb to pass testing

Additionally this drops testing of Python 3.8 and 3.9 which will be required by #333.

Closes #329


.github/workflows/main.yml
```diff
@@ -359,6 +344,104 @@ jobs:
...
+    needs:
+      [
+        check-changes,
+        prepare-nonhindcast-cache,
+        prepare-wave-hindcast-cache,
+        prepare-wind-hindcast-cache,
+      ]
+    if: |
+      always() &&
+      (
```
akeeste, line 358: "always() &&" seems redundant

.github/workflows/main.yml
```diff
@@ -359,6 +344,104 @@ jobs:
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.11'
+
+      - name: Generate matrix
+        id: set-matrix
+        run: |
+          matrix_json=$(python .github/workflows/generate_notebook_matrix.py)
```
akeeste, line 386: What is the purpose of the ``matrix_json`` variable here? Isn't ``matrix`` the only variable from ``generate_notebook_matrix.py`` that is used? Similarly, where is the ``matrix`` key "include" being used?

.github/workflows/generate_notebook_matrix.py
```diff
@@ -0,0 +1,56 @@
...
+    "default": 300,  # Default timeout for other notebooks
+}
+notebooks = []
+for root, dirs, files in os.walk("examples"):
+    for file in files:
+        if file.endswith(".ipynb"):
+            notebooks.append(os.path.join(root, file))
+
+# Generate the matrix configuration
+matrix = {"include": []}
```
akeeste, line 45: Does the key "include" have to do with how Actions creates a testing matrix from the test variable?

akeeste (approved):
Thanks @ssolson 

This all looks good to me. All the notebook tests are all running. I have a few clarifying questions to understand the Actions set-up better, but no issues with the approach.

.github/workflows/main.yml
```diff
@@ -359,6 +344,104 @@ jobs:
...
+          echo "source ~/miniconda3/etc/profile.d/conda.sh" >> ~/.bashrc
+          echo "conda activate TESTconda" >> ~/.bashrc
+          source ~/.bashrc
+
+      - name: Run notebook
+        shell: bash -l {0}
+        run: |
+          if [[ "${{ matrix.notebook }}" == "examples/metocean_example.ipynb" || "${{ matrix.notebook }}" == "examples/WPTO_hindcast_example.ipynb" ]]; then
+            if [[ "${{ needs.check-changes.outputs.should-run-hindcast }}" == 'true' ]]; then
+              jupyter nbconvert --to notebook --execute --inplace --ExecutePreprocessor.timeout=${{ matrix.timeout }} "${{ matrix.notebook }}"
```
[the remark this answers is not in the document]
ssolson: Yes each notebook in the repo has a custom timeout specified. It will fail if you go overtime. You can see some of these failures  on previous commits such as the commit before "Increase ADCP timeout".

.github/workflows/main.yml
ssolson: I agree, however, I have become accustomed to this syntax in actions and you will find it throughout our tests. 

There is discussion here agreeing that you should not need to but that you do:
https://github.com/orgs/community/discussions/25789#discussioncomment-3249273

.github/workflows/main.yml
ssolson: In bash here we are setting a variable `matrix_json` 
On the next line we echo that variable to send it as std_out via ">>" to the variable `matrix`.

So the `matrix_json` variable is used in the echo command

.github/workflows/generate_notebook_matrix.py
ssolson: Yes. 

The include key is a dummy key/ catch all for Actions because you can only have a single variable json at the top level. 
So its called include to include all the keys I want to use.

PR #309, opened 2024-04-24
In `wave.resource.surface_elevation` use `sum_of_sines` method if input spectrum does not have a zero frequency

simmsa:
Add `auto` method to wave.resource.surface_elevation to address #308. 

The `auto` method chooses the most computationally efficient surface
elevation computation method based on the input spectrum. `auto` uses
the `sum_of_sines` method is a zero frequency is not defined in the input spectrum
index and uses the `ifft` for all other cases.

PR #340, opened 2024-07-09
Improve `surface_elevation` internal method selection

simmsa:
Updated the `surface_elevation` calculation method to default to `sum_of_sines` when zero frequency is absent in the input spectrum. A warning is issued to inform users of this change. This ensures users benefit from the fastest computation method without needing to understand the underlying calculations.


akeeste (approved):
Thanks @simmsa. We previously addressed my one comment on approach in #309. No further changes from me

PR #336, opened 2024-06-27
Add optional parameter frequency_dimension across wave.resource

akeeste:
This PR makes the xarray work in wave.resource more consistent by adding the optional parameter ``frequency_dimension`` to functions like ``significant_wave_height`` that then pass the parameter to ``frequency_moment``

PR #341, opened 2024-07-09
Fix: Handle unnamed index/dimension in wave.resource.surface_elevation 

simmsa:
Fix: If unnamed, set spectrum dim_0 name to "Frequency"

The surface_elevation function typically takes in a spectrum generated
by an mhkit function that sets the index/dimension name to "Frequency".
This fix handles an atypical case where the index/first dimension is
unnamed, which causes downstream issues where we are expecting a
dimension named "Frequency".

Per the [xarray DataArray documentation](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.html)
an unnamed dimension starts at dim_0. If we find this case we
change the dim_0 name to "Frequency".

### DIFF
--- a/mhkit/tests/wave/test_resource_spectrum.py
+++ b/mhkit/tests/wave/test_resource_spectrum.py
@@ -148,6 +148,17 @@ class TestResourceSpectrum(unittest.TestCase):
 
         self.assertLess(rmse_sum, 0.02)
 
+    def test_spectrum_without_frequency_index_name_defined(self):
+        S = wave.resource.jonswap_spectrum(self.f, self.Tp, self.Hs)
+        S.index.name = None
+
+        eta_ifft = wave.resource.surface_elevation(S, self.t, seed=1, method="ifft")
+        eta_sos = wave.resource.surface_elevation(
+            S, self.t, seed=1, method="sum_of_sines"
+        )
+
+        assert_allclose(eta_ifft, eta_sos)
+
     def test_ifft_sum_of_sines(self):
         S = wave.resource.jonswap_spectrum(self.f, self.Tp, self.Hs)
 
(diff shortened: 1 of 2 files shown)


akeeste:
@simmsa Can you describe further the case that causes this issue? I can see how this will be a problem, but ``surface_elevation`` does not return any info with a frequency dimension, so renaming it here will not fix the downstream workflow.

Is there another ``wave.resource`` function that creates but does not properly name the frequency dimension? In #336 I propagated the variable ``frequency_dimension`` further so that we always follow the user input.

akeeste (approved):
Thanks @simmsa for catching this. This PR looks good to me. Were there any more changes to add before this is merged?

akeeste:
@simmsa Sorry about the delay here. Based on one of our conversations I wasn't sure if this was 100% done. It looks good as is. I'll merge this into the develop branch after v0.8.2 is released.

simmsa:
@akeeste, no worries, it looks like I didn't follow up with you here.

This fixes a potential bug where a user inputs `S` without frequency index named "Frequency" by using the found or defined `frequency_dimension` variable: https://github.com/MHKiT-Software/MHKiT-Python/blob/406599e3129c1090d522dc806a4ac0493a360f71/mhkit/wave/resource.py#L303-L304 to define names of the DataArray dimension and coordinates. 

Without this fix, passing a DataFrame with an unnamed index yields: 

`ValueError: coordinate Frequency has dimensions ('dim_0',), but these are not a subset of the DataArray dimensions ['Time', 'Frequency']`

The tests verify that both computation methods, `ifft` and `sum_of_sines` work as expected, and verify that a user defined dataframe also works as expected.

Happy to clarify further if necessary.

PR #348, opened 2024-08-15
Improve type handling and resolve #339

### LINKED ISSUE #339: Wave performance `mean_annual_energy_production_matrix` has no test & is incompatible with example
In the `wave_example.ipynb` Input box [15] calls:

```
# Calcaulte maep from matrix
maep_matrix = wave.performance.mean_annual_energy_production_matrix(
    LM_mean, JM, LM_freq
)
```

In MHKiT v0.8.1 this will fail because `mean_annual_energy_production_matrix` is not setup to handle 2D DataFrames. 
Additionally the check:
```
    if not np.abs(frequency.sum() - 1) < 1e-6:
        raise ValueError("Frequency components must sum to one.")
```         
was added but does not work for this LM_freq matrix provided.

@akeeste could you add a test for this function and determine why the example `LM_freq` does not pass this check which was previously commented out.

https://github.com/MHKiT-Software/MHKiT-Python/blob/a07ecc427ab924e32f8a2305d4be7193e5125614/mhkit/wave/performance.py#L258

For #330 I will calculate manually and leave the MHKiT function call commented out.

akeeste:
This PR resolves #339. It makes some improvements to the type handling, especially in the wave module, by allowing convert_to_dataarray to take in multivariate xr.datasets and pd.dataframes

Another larger PR to better utilize this improved type handling in the wave module will come after.


akeeste:
@ssolson I added wrapped up the last things on this PR:
- added the requirement on xarray dataset inputs to the docstring
- recapping our meeting discussion here for reference: the ``convert_to_dataarray`` parameter ``name`` is an attribute of the DataArray. It is the name of the variable that the DataArray represents. If the DataArray is converted to a Dataset, it automatically becomes the label of that data in the Dataset. This is different than the "variable" dimension which is created when a multivariate Dataset is converted to a DataArray.
- confirmed that the wave_example.ipynb now works as intended

akeeste:
@ssolson tests are passing and all review comments were addressed. I'll merge and continue working to speed up xarray inputs in the wave module

ssolson (approved):

PR #333, opened 2024-06-26
Numpy 2.0

ssolson:
Make MHKiT compatible with numpy 2.0

PR #351, opened 2024-09-04
Python 3.12

ssolson:
Add Python 3.12 support


akeeste:
Thanks @ssolson 

Is ``setuptools`` only required for the test environment or for MHKiT generally using Python 3.12? If the former, let's merge this now. If the latter, let's wait until our next release is closer and see if h5pyd has this fix on pypi then

akeeste (approved):

ssolson:
> Thanks @ssolson
> 
> Is `setuptools` only required for the test environment or for MHKiT generally using Python 3.12? If the former, let's merge this now. If the latter, let's wait until our next release is closer and see if h5pyd has this fix on pypi then

It is the latter, however, as this is the development branch I think developing all PRs for 3.12 support now would be beneficial rather than trying to fix any potential issues later.

ssolson:
I want to wait on #333 to be merged before this PR.

akeeste:
> It is the latter, however, as this is the development branch I think developing all PRs for 3.12 support now would be beneficial rather than trying to fix any potential issues later.

Fair point, that makes sense to avoid a bunch of urgent bugfixes right before we want to tag a release. Understood on timing this relative to #333. Tests pass on 3.12.

I successfully created a 3.12 environment with MHKiT locally, but FYI I did get a ``pip`` deprecation warning when using the editable install (``pip install -e .``) This method will not be supported in pip>=25.0 and will require another method

PR #357, opened 2024-10-14
Pylint Update &  Increase CDIP Example Time

ssolson:
The most recent update to the Pylint package added a warning for too many position arguments in addition to current too many arguments. We are currently ignoring historical too many arguments errors due to the major refactoring that is typically required to address. As such we will also ignore the too many positional arguments.

Additionally the CDIP API seems to be taking longer to respond. Local example runs are taking ~7 min.


akeeste (approved):
approved. Thanks @ssolson

Issue #355, opened 2024-09-25
PacWave clusters plot has flipped axes labels


jtgrasb:
### Describe the bug:
In the PacWave resource characterization example, the axes labels on the sea state representative clusters plot are flipped. 

### To Reproduce:
[PacWave Resource Characterization](https://mhkit-software.github.io/MHKiT/PacWave_resource_characterization_example.html)

**Minimal working example:** N/A

### Expected behavior:
x-axis should be period, y-axis should be height

### Screenshots:
![image](https://github.com/user-attachments/assets/184cadc0-b295-495d-b661-79e411ac9c06)

### Desktop (please complete the following information):
 - OS: N/A
 - MHKiT Version: 0.8.1

### Additional context:
N/A

ssolson:
I see. Good catch. I'll switch the labels around. 🙏

akeeste:
Resolved by #360

PR #360, opened 2024-11-21
Fix warnings in the documentation build

akeeste:
This PR pairs with https://github.com/MHKiT-Software/MHKiT/pull/80 and aims to eliminate the volume of warnings occurring in our documentation build.

Many of these changes will probably be fixing minor spacings/lines/issues in docstrings that show up as warnings but not errors when the documentation builds. Future linting should address and prevent these issues too.


akeeste:
@ssolson this should be a low lift to review when ready. We don't need to delay our release for this PR, but it would be great to merge whatever is pushed here and ready before the release. This will help me easy our documentation workflow and get that fully updated.

ssolson:
> @ssolson this should be a low lift to review when ready. We don't need to delay our release for this PR, but it would be great to merge whatever is pushed here and ready before the release. This will help ease future documentation updates and keep it fully updated.

No it should be `typing` and in default python I pushed a fix.

akeeste:
@ssolson I pushed a couple final commits to also resolve #355 as the pacwave notebook was resulting in an incorrect plot in the documentation.

ssolson (approved):
Thanks @akeeste this PR focuses on clean up of the code.

PR #361, opened 2024-11-27
MHKiT 0.9.0

ssolson:
# MHKiT v0.9.0 Release Notes

This release includes significant new features, enhancements, bug fixes, and updates to improve the functionality, performance, and compatibility of MHKiT.

### Acoustics Module (#349, #359)
- Introduced a new Acoustics module to MHKiT, providing comprehensive tools for processing and analyzing passive acoustic data from hydrophone recordings.
- Designed to facilitate compliance with the IEC-TS 62600-40 standard
- Supports hydrophone models such as SoundTrap and icListen, with functionality to read, calibrate, and process hydrophone data into acoustic metrics.
- Includes functionalities for:
  - Reading hydrophone data.
  - Performing spectral analyses.
    - spectral density (SPSD) 
    - spectral density levels (SPSDL) 
    - Fractional octave bands (e.g., third-octave, decidecade)
    - Sound Pressure Level (SPL)
  - Applying calibrations.
  - Calculating sound pressure levels.
  - Visualizing results.
  - Audio export
-  To support users, an example notebook demonstrates the application of this module with workflows.

### Strain Processing Example (#358)
The Strain Processing example introduces a workflow for analyzing strain gauge data, particularly from tidal turbine blade testing. While the functions are not included as a formal MHKiT module due to the variability of test setups, this example notebook provides a detailed and practical framework for processing such data.

- Illustrates the calibration and processing of strain gauge data, showcasing a real-world application in tidal turbine testing.
- Designed to accommodate the variability in strain gauge testing setups, enabling users to adapt the workflow to their specific needs.
- Includes figures and explanations to contextualize the analysis and support data interpretation.
- Reference:
  - Gunawan, B., Haulenbeek, K., Abdellatef, M., Streit, R., Lynn, E., Willis, M., Pendley, D., Gallegos-Patterson, D., Neary, V., Wosnik, M. (2024). Calibration Of Fiber Optic Rosette Sensors For Measuring Bending Moment On Tidal Turbine Blades. International Conference on Ocean Energy, Melbourne, Australia, September 17–19, 2024.

### Wave Resource Module Performance (#352)
Optimizes the wave.resource module to improve performance and resolve issue #331 by transitioning from xarray.Dataset to xarray.DataArray for core functions. Handling edge cases robustly in pure numpy proved challenging, so the adoption of DataArray strikes a balance between performance and usability.
- Transitioning to DataArray provides a significant speed-up—up to 1000x faster for very large datasets compared to the previous Dataset implementation.
[...]

PR #362, opened 2024-12-02
Fix Mac OS Tests 

ssolson:
DOLfYN tests are failing on macos. THis PR will fix this issue.

This first attempt normalizes the data to see if that reduces the noise in the comparison.


ssolson:
@jmcvey3 Mac OS fails on the DOLfYN `test_motion:test_motion_adv` test. (ubuntu and Windows both pass). See https://github.com/MHKiT-Software/MHKiT-Python/pull/361. 

image


I tried normalizing the data but that did not work so I reduced the absolute tolerance from 1e-7 to 1e-6 and it passes. 

Are you okay with this modification? Would you approach it differently?

PR #365, opened 2024-12-03
FIx Flow Duration Curve Plot Bug

### LINKED ISSUE #364: Flow duration curve
### Describe the bug:
There seems to be an issue with plotting the flow duration curve in the updated 0.8.x version. It was producing the expected plot in the 0.7.x version.

In particular, when using commit 2ccc286, I got the following plot

When using latest commit, I got 

### To Reproduce:
Sample script:

```
#!/usr/bin/env python3
#-----------------------------------------------------------------------#
from mhkit import river
#-----------------------------------------------------------------------#
# Obtain 5 years of daily discharge data
data = river.io.usgs.request_usgs_data(station="04216000",
                                       parameter='00060',
                                       start_date='2017-01-01',
                                       end_date='2022-01-31',
                                       data_type='Daily')

data = data.rename(columns={data.columns[0]: 'Q'})
data = data.reset_index()

# Convert to discharge data from ft3/s to m3/s
data.Q = data.Q/(3.28084)**3

# Calculate exceedence probability (flow duration curve)
data['ExcP'] = river.resource.exceedance_probability(data.Q)

# Plot FDC
fig = plt.axes()
river.graphics.plot_flow_duration_curve(data.Q, data.ExcP, ax=fig)
plt.show()
```

### Expected behavior:
There should be only one curve in the flow duration plot (discharge vs. exceedance probability).

### Desktop (please complete the following information):
 - OS: N/A
 - MHKiT Version: 0.8.x

ssolson:
Fixes #364 by assigning the sort by Calculation performed on the DataFrame.

image


akeeste (approved):
Thanks @ssolson for the quick fix. Works for me

Issue #366, opened 2024-12-04
DOLfYN issue reading large AD2CP files


akeeste:
### Describe the bug:
When using DOLfYN to read a large (12GB) Nortek Signature 1000 .ad2cp file, the initialization of ``_Ad2cpReader`` fails due to the file size. In this case, ``_Ad2cpReader._check_header()`` calls ``_reopen()`` with ``bufsize = self._eof`` , which is equivalent to the file size in bytes (~12e9). This value is too large for python to convert to a C int behind the scenes. The ``nens`` argument is not used here, and I cannot see any other user-facing arguments that can get around this issue.

I can manually decrease ``_eof`` during debugging and the rest of the ``read`` function continues as expected, though of course I don't know if the headers are being determined successfully by ``_check_header``. @jmcvey3 do you have a work around for files this size?

### To Reproduce:
@akeeste can send a link to the file in question if needed

**Minimal working example:**
Using the develop branch:
```
import mhkit.dolfyn as dolfyn
ds_adp = dolfyn.io.api.read("S100604A032_5_beam_8Hz.ad2cp", nens=[0,100])
```

### Expected behavior:
Expected large file size to not be an issue when nens is used.

### Screenshots:
Error messages:
```
  Traceback (most recent call last):
    File "c:\Users\akeeste\anaconda3\envs\mhkit\Lib\runpy.py", line 198, in _run_module_as_main
      return _run_code(code, main_globals, None,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "c:\Users\akeeste\anaconda3\envs\mhkit\Lib\runpy.py", line 88, in _run_code
      exec(code, run_globals)
    File "c:\Users\akeeste\.vscode\extensions\ms-python.debugpy-2024.0.0-win32-x64\bundled\libs\debugpy\adapter/../..\debugpy\launcher/../..\debugpy\__main__.py", line 39, in <module>
      cli.main()
    File "c:\Users\akeeste\.vscode\extensions\ms-python.debugpy-2024.0.0-win32-x64\bundled\libs\debugpy\adapter/../..\debugpy\launcher/../..\debugpy/..\debugpy\server\cli.py", line 430, in main
      run()
[...]

PR #371, opened 2025-01-09
Avoid failing to scan very large files

jmcvey3:
Simple fix to a problem created by a previous PR. Solves Issue #366


akeeste:
Thanks @jmcvey3! Looks good to me and all tests pass. Merging

Issue #370, opened 2025-01-07
DOLfYN reading a .VEC file indefinitely


akeeste:
### Describe the bug:
@jmcvey3 I'm assisting another user with reading their .VEC file. DOLfYN gets stuck and reads it indefinitely. Other VEC files from this user are read normally.

The issue occurs when ``_NortekReader.init_ADV()`` is attempting to determine the [code spacing](https://github.com/MHKiT-Software/MHKiT-Python/blob/bc1eadc9117f969bed618a104834892bff1e3df0/mhkit/dolfyn/io/nortek.py#L308). In ``code_spacing``, the first call to ``self.findnextid()`` gets stuck in the [while loop](https://github.com/MHKiT-Software/MHKiT-Python/blob/bc1eadc9117f969bed618a104834892bff1e3df0/mhkit/dolfyn/io/nortek.py#L458), seemingly because the ID it is searching for is never found.

### To Reproduce:
``ds_adv2 = dolfyn.io.api.read("SANDIA02.VEC")``. @akeeste can share the data directly if needed.

### Expected behavior:
Run without error, as on other VEC files

### Desktop (please complete the following information):
 - OS: Windows 11
 - MHKiT Version: v0.9.0

jmcvey3:
File appeared to be either corrupted or filled with "fill" bytes added during real-time operation. There were two "while loops" that the script became hung up on, and PR #372 added logic to escape these loops.

akeeste:
Resolvd by #372

PR #372, opened 2025-01-09
Fix for corrupted Nortek files

jmcvey3:
If a datafile is filled with bad bytes, the reader will get stuck in one of two while loops.

### DIFF
--- a/mhkit/dolfyn/io/base.py
+++ b/mhkit/dolfyn/io/base.py
@@ -83,6 +83,9 @@ def _handle_nan(data):
     Finds trailing nan's that cause issues in running the rotation
     algorithms and deletes them.
     """
+    if "time" not in data["coords"]:
+        raise Exception("No data recorded in file.")
+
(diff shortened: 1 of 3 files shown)


akeeste:
@jmcvey3 Using the sample file I sent you, this fix gets past the previous sticking point in #370 (``_NortekReader`` initialization), but now gets stuck at ``_NortekReader.readfile()``. Is there another check you can add that will break the while loops in ``_NortekReader.readfile()`` and/or ``_NortekReader.findnext()``?

jmcvey3:
> @jmcvey3 Using the sample file I sent you, this fix gets past the previous sticking point in #370 (`_NortekReader` initialization), but now gets stuck at `_NortekReader.readfile()`. Is there another check you can add that will break the while loops in `_NortekReader.readfile()` and/or `_NortekReader.findnext()`?

How long did you let it run for, and what is listed in the .dolfyn.log? It took a good 5 or 10 minutes for dolfyn to cycle through every byte of the file to try and find valid data.

PR #376, opened 2025-02-06
Update integral length scale function

jmcvey3:
Discovered a problem with the original function while conducting data analysis of ADV data from Puget Sound. 

Updating this function per the textbook definition of the integral time scale - it is listed as Equation 5.3 in this paper (http://dx.doi.org/10.1098/rsta.2012.0196). The code I'm replacing defined the integral time scale as when the autocorrelation function decays to 1/e of its original value, which is less robust than finding the first zero-crossing of the autocorrelation function, as is standard. 

The integral length scale [m] is simply the average flow speed [m/s] multiplied by the integral time scale [s].


ssolson (commented):
@jmcvey3 theses changes are good to go once the tests pass.

I saw you started all 4 PR checks at the same time yesterday which will cause the APIs to fail.

I restarted just this one for now and hope it will pass and we can get this merged here in the next couple hours.

Issue #373, opened 2025-01-24
Dual profile naming scheme and running (dolfyn.adp) cleaning functions


willcoxe:
Hello,

I am using the dual profile option in dolfyn to read an ad2cp file and then attempting to perform some basic operations according to 
[the adcp example](https://github.com/MHKiT-Software/MHKiT-Python/blob/main/examples/adcp_example.ipynb). For dual profiles sda and sdb, sda functions perfectly but because the names are appended with an _avg in sdb, the second profile, many of the operations do not run correctly, such as rotating the frame and/or cleaning data based on pressure & on correlation. Also in the second profile the dsb.attrs["cell_size"] is missing which leads to issues with some of the operations described in the example. 

May of these work once the _avg is removed from the name and the time dimension is adjusted to dsb["time"] instead of dsb["time_avg"] (e.g. ```dsb = dsb.swap_dims({'time_avg': 'time'})``` ) and I make sure to manually set dsb.attrs["cell_size"]. 

One exception (there might be others), which I have not yet fully debugged is api.clean.remove_surface_interference(dsb). 

```
>>> api.clean.remove_surface_interference(dsb)
Traceback (most recent call last):
  File "~/code/python/venvs/adcp2/lib/python3.13/site-packages/mhkit/dolfyn/adp/clean.py", line 397, in remove_surface_interference
    a[..., bds] = val
    ~^^^^^^^^^^
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "myscript.py", line 95, in <module>
    dsb = api.clean.remove_surface_interference(dsb)
  File "~/code/python/venvs/adcp2/lib/python3.13/site-packages/mhkit/dolfyn/adp/clean.py", line 399, in remove_surface_interference
    a[..., bds] = 0
    ~^^^^^^^^^^
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed

```

I am guessing that others may run into this also when using dual profiles with Nortek instruments. I will continue to try to debug the problem and would appreciate any hints you might have.

willcoxe:
It was 'range'. I have added ```dsb = dsb.swap_dims({'range_avg': 'range'})``` and it works now. 

Might still be an idea to somehow catch this in the code somewhere.

ssolson:
Thanks @willcoxe for letting us know your experience. 

[...]

PR #377, opened 2025-02-06
Allow clean functions to handle _avg variables

jmcvey3:
Makes the ADCP cleaning functions more robust - updated based on latest reader updates for dual profiling instruments. Solution for Issue #373


mhkit/dolfyn/adp/clean.py
```diff
@@ -157,30 +157,38 @@ def water_depth_from_amplitude(ds, thresh=10, nfilt=None) -> None:
...
     D = np.vstack((d1, d2))
     # Take the median value as the estimate of the surface:
     d = np.median(D, axis=0)
 
     # Throw out values that do not increase near the surface by *thresh*
-    for ip in range(ds["vel"].shape[1]):
+    for ip in range(ds["vel" + tag[0]].shape[1]):
         itmp = np.min(inds[:, ip])
         if (edf[itmp:, :, ip] < thresh).all():
             d[ip] = np.nan
```
ssolson, line 194: on 188 here you are using the median on d1 and d2 and I notice on 194 you add a nan to the array.

Will d1 or d2 ever have nan? If so median will always return nan. Is that the behavior you want? Or would `np.nanmedian` be preferred?

mhkit/dolfyn/adp/clean.py
```diff
@@ -259,16 +267,26 @@ def water_depth_from_pressure(ds, salinity=35) -> None:
...
+        for v in ds.data_vars
+        if (("temp" in v) and ("clock" not in v) and ("press" not in v))
+    ]
+    if not temp:
         raise NameError("The variable 'temp' does not exist.")
 
     # Density calcation
-    P = ds["pressure"].values
-    T = ds["temp"].values  # temperature, degC
+    P = ds[pressure[0]].values  # pressure, dbar
```
ssolson, line 288: do all instruments use dbar (over Pa)? or is it well described in the examples? Should this be added to the docstring?

mhkit/dolfyn/adp/clean.py
```diff
@@ -372,31 +390,47 @@ def remove_surface_interference(
...
+        a
+        for a in ds.attrs
+        if (
+            ("cell_size" in a)
+            and ("_bt" not in a)
+            and ("_alt" not in a)
+            and ("wave" not in a)
+        )
+    ]
+
```
ssolson, line 404: Should this: 
`raise KeyError("No valid 'cell_size' attribute found in dataset.")`

Or will there always be a "cell_size"?

ssolson (commented):
@jmcvey3 thanks for submitting this. I think most of these questions are for me but maybe I found something helpful.

mhkit/dolfyn/adp/clean.py
```diff
@@ -157,30 +157,38 @@ def water_depth_from_amplitude(ds, thresh=10, nfilt=None) -> None:
...
+    # Should not matter which is used.
+    tag = []
+    if hasattr(ds, "vel"):
+        tag += [""]
+    if hasattr(ds, "vel_avg"):
+        tag += ["_avg"]
+
     # This finds the maximum of the echo profile:
-    inds = np.argmax(ds["amp"].values, axis=1)
+    inds = np.argmax(ds["amp" + tag[0]].values, axis=1)
```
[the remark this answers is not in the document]
jmcvey3: Good thoughts; however, if "vel" doesn't exist, "amp" and "corr" will also not exist. The signal amplitude and correlation are the quality analysis of the velocity ping's signal.

mhkit/dolfyn/adp/clean.py
```diff
@@ -199,7 +207,7 @@ def water_depth_from_amplitude(ds, thresh=10, nfilt=None) -> None:
 
     ds["depth"] = xr.DataArray(
         d.astype("float32"),
-        dims=["time"],
+        dims=["time" + tag[0]],
```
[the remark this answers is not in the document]
jmcvey3: Yes, I hardcoded it. The data in the file stored under the "averaging" ID has their own timestamp, and I log "time" from that data ID with the "_avg" tag.

mhkit/dolfyn/adp/clean.py
jmcvey3: Hmmm this is a Levi function... I'll add nan to all of those median min and max functions, because if this is called after another QC function that would be a problem.

mhkit/dolfyn/adp/clean.py
jmcvey3: Yes, they all use dbar because it translates nearly 1:1 as meters-beneath-the-surface (assuming the pressure sensor was zeroed before deploying, of course)

mhkit/dolfyn/adp/clean.py
jmcvey3: There will always be a "cell_size", yes.

akeeste:
With Sterling's commits to `develop` pulled in, tests are now passing

akeeste:
@jmcvey3 are there any other outstanding changes in this PR from @ssolson's review?

jmcvey3:
> @jmcvey3 are there any other outstanding changes in this PR from @ssolson's review?

I was waiting on a response in issue #373, but looks like we're good to go.

Issue #368, opened 2024-12-19
DOLfYN Teledyne River depth


browniea:
The "river depth" output from WinRiver appears to have an offset or different reference that can be validated with the bathymetry data. 
I was looking in the user guide here for "river depth" [WinRiver II user guide](https://corr-tek.it/wp-content/uploads/2022/07/WinRiver-II-User-Guide_.pdf) There are a few different refences the river depth can be using. DOLfYN might not be using the right one. 
<img width="409" alt="image001" src="https://github.com/user-attachments/assets/ea371f18-ced7-4421-9f4f-8f6861eca28b" />

# WinRiver River Depth comparted to `mhkit.river.d3d`
This plot shows that the MHKiT river d3d module matches the output from WinRiver 
<img width="401" alt="image003" src="https://github.com/user-attachments/assets/91fbc073-094d-4960-bef1-1b3810f66ab0" />

# `mhkit.dolfyn` River Depth comparted to `mhkit.river.d3d`
This plot shows that the DOLfYN ADCP has an offset to the ADCP variable `dist_bt` read in by DOLfYN
<img width="410" alt="image004" src="https://github.com/user-attachments/assets/57ac5fbe-a5c7-4d97-8194-436f86e24860" />

I expect DOLfYN river D3D should match WinRiver and MHKiT in this case.

ssolson:
Close #375

jmcvey3:
PR #378 should officially solve this bug, which I didn't originally see, involving the changing depth cell sizes that the RiverPro/StreamPro firmware creates. I've removed RDI's "bin1_dist_m" and replaced it the dolfyn agnostic "range_offset", so the user knows how far the range coordinate is offset based on the ADCP's submerged depth.

PR #378, opened 2025-02-08
Fix ever-changing RDI RiverPro depth bin ranges

jmcvey3:
Partial solution for Issue #368. Fixes an issue with a changing `bin1_dist_m` attribute, which was presumed constant before. Also solves an issue with changing surface layer cell sizes. 
The RDI reader will now also explicitly return the `range_offset` that the user programmed into the instrument.


ssolson (approved):
Could you  add cfg to the docstring here?

I also need to get #380 in before the tests will pass on your PRs.

PR #379, opened 2025-02-12
Acoustics SPL bugfix

jmcvey3:
Fixes a bug where calculating SPLs in lower frequency bands was failing because the bandwidth resolution was on par with the frequency resolution, i.e. numpy.trapz was failing to calculate the integral because the input was only 1D over frequency. Also expanded testing to cover this particular case.


mhkit/acoustics/analysis.py
```diff
@@ -795,28 +817,27 @@ def _band_sound_pressure_level(
...
-            spsd["freq"].sel(freq=slice(band_min, band_max)),
-        )
+        # Min and max band limits
+        band_range = [octave_bins[i], octave_bins[i + 1]]
+
+        # Interpolate between band frequencies if width is narrow
+        x = spsd["freq"].sel(freq=slice(*band_range))
+        if len(x) < 2:
+            spsd_slc = spsd.interp(freq=band_range)
+            x = band_range
```
ssolson, line 835: If the dataset is large could this interp lead to performance issues where we would want to warn the user of what was slowing down the code?

mhkit/acoustics/analysis.py
```diff
@@ -473,6 +473,44 @@ def _validate_method(
     return method_name, method_arg
 
 
+def _create_frequency_bands(octave, fmin, fmax):
```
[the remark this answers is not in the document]
jmcvey3: I didn't add these because they're already checked in the respective functions that use this one

mhkit/acoustics/analysis.py
jmcvey3: Good question, I'm not sure.

This if statement should only happen at low frequencies, so in the next commit I'll slice the dataset down before interpolating. Should be faster.

Issue #275, opened 2023-11-07
Automatic Linting


ssolson:
Enforce a style guide on files in the MHKiT repository via GitHub Actions.

ssolson:
I am going to break this task into two PRs:
- auto formatting via black (#281)
- linting via pylint

PR #389, opened 2025-04-01
Lint river module

ssolson:
This PR adds pylint enforcement to the MHKiT river module. 
Additionally this PR adds type hints to the river module functions.
Part of #275


mhkit/river/performance.py
```diff
@@ -1,7 +1,16 @@
+"""
+performance.py
+
+Computes device metrics such as equivalent diameter, tip speed ratio,
+and capture area. Calculations are based on IEC TS 62600-300:2019 ED1.
+
+"""
```
akeeste, line 7: This is the equivalent description in our docs that will be replaced by the performance module docstring. Is any of this information necessary here? Seems like the sentences clarifying turbine types could move to each function's doc string

"The performance submodule contains functions to compute equivalent diameter and capture area for circular, ducted, rectangular, adn multiple circular devices. A circular device is a vertical axis water turbine (VAWT). A rectangular device is a horizontal axis water turbine. A ducted device is an enclosed VAWT. A multiple-circular devices is a device with multiple VAWTs per device. The performance module also includes functions for calculating a turbine coeffcient of power and tip speed ratio."

mhkit/river/resource.py
```diff
@@ -1,11 +1,24 @@
+"""
+resource.py
+
+Computes resource assessment metrics, including exceedance probability,
+inflow velocity, and power (theoretical resource). Calculations are based
+on IEC TS 62600-301:2019 ED1.
+
+"""
```
akeeste, line 8: Noting below the equivalent river.resource module docstring currently in the documentation.  The new description looks good to me.

"The resource submodule uses discharge data to compute exeedance probability, velocity, and power. The module also contains functions to compute the Froude number and to fit a polynomial to a series of points. The polynomial is used to estimate the relationship between discharge and velocity or velocity and power at an individual turbine."

mhkit/river/graphics.py
```diff
@@ -1,10 +1,31 @@
+"""
+graphics.py
+
+The graphics module provides plotting utilities for river energy resource data.
```
akeeste, line 4:
```suggestion
The graphics module provides plotting utilities for river energy resource data.

File: mhkit/river/io/__init__.py [hunk 4]
```diff
@@ -1,2 +1,9 @@
+"""
+river.io
+
+This module provides input/output functionality for river energy related data in MHKiT.
```
akeeste, line 4:
```suggestion
This module provides input/output functionality for river energy related data in MHKiT.

File: mhkit/river/io/d3d.py [hunk 5]
```diff
@@ -1,13 +1,24 @@
-from mhkit.utils import unorm
-import scipy.interpolate as interp
+"""
+d3d.py
+
+This module provides functions for reading and processing Delft3D (D3D) model output data.
```
akeeste, line 4:
```suggestion
This module provides functions for reading and processing Delft3D (D3D) model output data.

File: mhkit/river/io/usgs.py [hunk 6]
```diff
@@ -1,12 +1,39 @@
+"""
+usgs.py
+
+This module provides functions for retrieving and processing data from the United States
```
akeeste, line 4:
```suggestion
This module provides functions for retrieving and processing data from the United States

File: mhkit/river/performance.py [hunk 7]
```diff
@@ -1,7 +1,16 @@
...
+performance.py
+
+Computes device metrics such as equivalent diameter, tip speed ratio,
```
akeeste, line 4:
```suggestion
Computes device metrics such as equivalent diameter, tip speed ratio,

File: mhkit/river/resource.py [hunk 8]
```diff
@@ -1,11 +1,24 @@
...
+resource.py
+
+Computes resource assessment metrics, including exceedance probability,
```
akeeste, line 4:
```suggestion
Computes resource assessment metrics, including exceedance probability,

**Review by akeeste — COMMENTED** [review]
Thanks @ssolson, this looks great!

My one minor request is that we remove the module names from top-level docstrings, otherwise the API docs will format a bit strangely (the duplicate description in the image would be removed when the API docs are updated). I added suggestions where I caught them. 
 
image

Are there any functional changes in the two notebooks?

File: mhkit/river/performance.py (hunk 1 above)
Comment (ssolson) [reply mhkit/river/performance.py:7]: Yes this information should be here including the functions although I have removed them short term because I decided to punt the discussion until later so we could have a strategy around the autodocs.

Here is the standard: https://peps.python.org/pep-0257/#multi-line-docstrings

> The docstring for a module should generally list the classes, exceptions and functions (and any other objects) that are exported by the module, with a one-line summary of each. (These summaries generally give less detail than the summary line in the object's docstring.) The docstring for a package (i.e., the docstring of the package's __init__.py module) should also list the modules and subpackages exported by the package.


This is useful because a user should be able to call help on a module and know what functions are available. 
E.g:
```
>>> import x; help(x)
```
Should inform the user what they have imported and what functions are available.

ssolson:
> Thanks @ssolson, this looks great!
> 
> My one minor request is that we remove the module names from top-level docstrings, otherwise the API docs will format a bit strangely (the duplicate description in the image would be removed when the API docs are updated). I added suggestions where I caught them.

Done.

> Are there any functional changes in the two notebooks?

No functional notebook changes.

akeeste (approved):
Thanks @ssolson this looks good to me

PR #393, opened 2025-04-28
MLER explanation updates/corrections

rgcoe:
I looked at the descriptive text on the MLER example and noticed some problems. I've made some quick corrections in this PR.


akeeste (approved):
Thanks @rgcoe, approved and merging

PR #395, opened 2025-04-29
Add WEC-Sim power performance example 

akeeste:
This PR adds an example that uses MHKiT and WEC-Sim together to do a power performance assessment:
- utilize resource characterization in PacWave notebook
- write MCR cases for WEC-Sim
- load WEC performance
- assess WEC performance

Right now I represent the resource using the 32-cluster results in the PacWave notebook and irregular waves in WEC-Sim. I was considering writing all 2000+ wave cases in that example to a WEC-Sim MCR file so that we can demonstrate `mhkit.wave.performance.power_performance_workflow()` but decided it would be too expensive, even if regular waves were used in WEC-Sim.

I also have a half-finished example that shows how to write a directional spectra (like in `directional_waves.ipynb`) for WEC-Sim. But the WEC-Sim 'full directional spectra" feature needs documentation and an example first. 

The PR also includes a function to make the xarray representation of WEC-Sim data a lot more useful


MShabara:
I’ve tested the WEC-Sim examples with a variety of WEC-Sim cases, including:

- Single-body problems  
- Two-body systems  
- Moored and unmoored WECs  

Overall, the implementation works well across these scenarios. However, I’ve noticed a few points worth mentioning:

- **Wave spectrum capture**: The wave spectrum is not always captured correctly. The accuracy appears to be sensitive to the tuning parameters, particularly the number of lines and the sampling rate.
- **Flexible body compatibility**: The current implementation does not support flexible bodies, as the `body` class for flexible components differs in structure. This could be addressed in future development.

These limitations are not critical for merging but are worth noting for future improvements.

Great work on this feature!

If my final commit works for you, feel free to merge the PR

examples/wecsim_power_performance_example.ipynb
MShabara, line 1: A commit was made to this file. Two cosmetic changes.

akeeste:
@MShabara understood on GBM as that has an entirely different output structure.--this can be a separate future PR. 

On the wave spectrum--where did you see this inconsistency coming in? Is it based in WEC-Sim or MHKiT?

Issue #392, opened 2025-04-25
Read issue with Signature ADCP dataset


hevgyrt:
### Describe the bug:
I get an OverflowError when reading a subset (using the `nens` argument) of a large (>10Gb) Signature 500 .ad2cp file. The adcp was measuring in a concurrent mode including the "burst and waves" + the echosounder option.

### To Reproduce:
Provide a minimal working example that reproduces the error. 

`ds = dolfyn.read('my_sig500_file.ad2cp',nens=[2000,20000])`

prompts

`File ~/miniconda3/lib/python3.10/site-packages/mhkit/dolfyn/io/api.py:114, in read(fname, userdata, nens, **kwargs)
    112     func_map = dict(RDI=read_rdi, nortek=read_nortek, signature=read_signature)
    113     func = func_map[file_type]
--> 114 return func(fname, userdata=userdata, nens=nens, **kwargs)

File ~/miniconda3/lib/python3.10/site-packages/mhkit/dolfyn/io/nortek2.py:79, in read_signature(filename, userdata, nens, rebuild_index, debug, dual_profile, **kwargs)
     75             raise TypeError("nens must be: None (), int, or len 2")
     77 userdata = _find_userdata(filename, userdata)
---> 79 rdr = _Ad2cpReader(
     80     filename, rebuild_index=rebuild_index, debug=debug, dual_profile=dual_profile
     81 )
     82 d = rdr.readfile(nens[0], nens[1])
     83 rdr.sci_data(d)

File ~/miniconda3/lib/python3.10/site-packages/mhkit/dolfyn/io/nortek2.py:157, in _Ad2cpReader.__init__(self, fname, endian, bufsize, rebuild_index, debug, dual_profile)
    155 self.f.seek(0, 2)  # Seek to end
    156 self._eof = self.f.tell()
--> 157 self.start_pos = self._check_header()
    158 self._index, self._dp = lib.get_index(
    159     fname,
    160     pos=self.start_pos,
   (...)
    164     dp=dual_profile,
    165 )
    166 self._reopen(bufsize)

File ~/miniconda3/lib/python3.10/site-packages/mhkit/dolfyn/io/nortek2.py:209, in _Ad2cpReader._check_header(self)
    206         idx = s.find(c, idx + 1)
    208 # Open the entire file
--> 209 self._reopen(self._eof)
    210 pk = self.f.peek(1)
    211 # Search for multiple saved headers

[...]

Issue #391, opened 2025-04-22
Reading Signature 100 ad2cp file not working


jkolbu:
Hi! Have several deployments of deep ADCP data to process, however the files dont want to open with 'orientmat_avg' being the problem - despite it being set to 'AUTOZUPDOWN'

The file size is large but I can upload whatever files are most useful. 

![Image](https://github.com/user-attachments/assets/59d19496-21e8-4f87-aec9-7139a223e391)

ssolson:
Hey @jkolbu thanks for your interest in MHKiT and bringing this to our attention. 

@jmcvey3  is our Dolfyn expert and hopefully he can weigh in or help soon.  If its possible to upload a down sampled version of the data creatnig the error above that would be helpful. If not can you share the full data here or share it via google drive if its too large?

jkolbu:
Hi @ssolson thanks for the quick reply - see google drive link here - both sets of data in the two folders are giving the same error - https://drive.google.com/drive/folders/1mIb--LGBv3Crcf0nSI2fAuTekSgOR5N-?usp=sharing

jmcvey3:
Hi @jkolbu, thanks for bringing up the issue! We haven't seen a Sig100 file before. I was able to fix the issue you outline here pretty quickly and can the read the entire "S106939A004_ANTLS12DANA.ad2cp" file.

Was the "S106939A007_SH2_DA_2400.ad2cp" mentioned in the deployment log in folder SH2_DA altered to get the "S106939A007_SH2_DA_2400.ad2cp" and "S106939A007_SH2_DA_2400_avg.ad2cp" files? Dolfyn is skipping a lot of what it thinks are corrupted bytes in those two files.

jkolbu:
Hm interesting - no it wasn't - these were downloaded directly from the Signature software.

jmcvey3:
@jkolbu If you pull PR #397, you should now be able to read the three .ad2cp files you shared. The larger files with take some minutes each depending on your computer's processing power.

jkolbu:
Thanks very much - working now!

jmcvey3:
> Thanks very much - working now!
> […](#)

@jkolbu Glad to hear it! I just pushed one more commit that should save the duty cycling parameters in the dataset attributes, as well as properly log the sampling frequency.

PR #397, opened 2025-05-05
Improve Nortek2 index file creator functions

jmcvey3:
Improves the nortek2 reader to handle block headers with a length of 12 bytes (which can happen for large data blocks according to the Nortek integrator guide pdf - looking at the echo sounder here).

Improves/makes more robust the ability to search for a second global configuration header if a file is corrupted/rewritten/appended to for various edge cases.

Adds ability to read AD2CPs that are dual profiling with an echo sounder, though the dual-profiling functionality is still effectively in "beta".

Solution for Issue #391 and an update to PR #371 (which should have solved Issue #392).


akeeste (approved):
Thanks @jmcvey3 . As we discussed this morning, it's pretty hard to review these reader PRs effectively, but the user's issue in #391 was solved by this PR and both current and new tests are passing.

ssolson (approved):

PR #398, opened 2025-05-16
Lint hindcast

ssolson:
This PR adds type hints to functions and will enforce linting on the wave hindcast module. 

I want to separate out the wave hindcast from the remainder of the wave module due to the long testing times on the hindcast module.


akeeste (approved):
Thanks @ssolson, this looks good to me

PR #401, opened 2025-05-20
Configure specific warnings

ssolson:
This PR removes the ignore all future warnings added when some packages were suggesting updating to numpy 2.0 but mhkit needed other packages to update before that could happen.


akeeste:
Thanks @ssolson. I reviewed you work, and added a couple fixes.

One specific question for @jmcvey3 given the very particular in-line DOLfYN comment and then we can merge this PR

PR #405, opened 2025-06-27
Short list of VMDAS updates

jmcvey3:
Discovered from a recent vessel survey that I misunderstood TDI's documentation on a couple of VMDAS parameters. The speed and direction parameters I thought it was pulling directly from the GPS it is calculating from the change in latitude/longitude over time. I renamed them and added the two variables that are actually directly from the GPS.

Also discovered that VMDAS cannot take an RMC NMEA sentence, so the GPS timestamps are lacking a datestamp. Which is not helpful if your ADCP's clock is on local time and the GPS is always in UTC. So I added a code block that checks if the GPS UTC time rolled over from 23:59:59 to 00:00:00 and manually updated the day.


akeeste (approved):
Thanks @jmcvey3 . One minor question on how frequently the day-skipping could happen

jmcvey3:
> Thanks @jmcvey3 . One minor question on how frequently the day-skipping could happen

It's a VMDAS software issue - all of the NMEA0183 sentences record the timestamp in UTC, but not the date. (Only the RMC sentence records the date.) When VMDAS saves the timestamps, it saves them in order, but using the date from the laptop/ADCP. This means the date will not change, and the time will roll over from 23:00 to 00:00. 
They're essentially assuming the work laptop/ADCP are working in UTC time, which if they are this wouldn't be a problem.

PR #406, opened 2025-07-09
Allow user to specify universal Kolmogorov constant for TKE dissipation rate function

jmcvey3:
It was pointed out to me that a universal constant of 0.5 was hardcoded into dolfyn when calculating dissipation rate using the Lumley and Terray method. According to turbulence theory and historical experimental validation, this is the value to use for the streamwise velocity spectra. According to the same, the constant is different for velocity spectra that are orthogonal to the streamwise direction - usually taken to be somewhere between 0.65 and 0.69.

This PR allows the user to set this constant via function input.

Pope, Stephen B. "Turbulent flows." Measurement Science and Technology 12.11 (2001): 2020-2021. pp 231-232.


akeeste (commented):
Thanks @jmcvey3, this looks great. One minor comment, line 439 can break if `k_constant` is a float and the psd has all 3 velocity components:
```
k_constant = 0.5
a = np.array(k_constant)
a = a[:, None, None]
```

instead of an iterable of size 1:
```
k_constant = [0.5]
a = np.array(k_constant)
a = a[:, None, None]
```

a couple options:
- force k_constant input to be iterable(1) or iterable(3), or
- convert a k_constant float to iterable(1)

jmcvey3:
> Thanks @jmcvey3, this looks great. One minor comment, line 439 can break if `k_constant` is a float and the psd has all 3 velocity components:
> 
> ```
> k_constant = 0.5
> a = np.array(k_constant)
> a = a[:, None, None]
> ```
> 
> instead of an iterable of size 1:
> 
> ```
> k_constant = [0.5]
> a = np.array(k_constant)
> a = a[:, None, None]
> ```
> 
> a couple options:
> 
> * force k_constant input to be iterable(1) or iterable(3), or
> * convert a k_constant float to iterable(1)

Thanks for the look @akeeste, I'll update with a solution

PR #409, opened 2025-08-04
DOLFyN/RDI: Set  `fs` to NaN when typical calculation methods yield error (#408)

### LINKED ISSUE #408: Problem reading RDI Pinnacle 45 burst data with read_rdi
Describe the bug:

I get the following error message when I try to read a pd0-file from a RDI Pinnacle 45 instrument using read_rdi with mhkit version 0.9.0:

```
ZeroDivisionError                         Traceback (most recent call last)
Cell In[3], line 1
----> 1 data = read_rdi(file)

File [...]/MHKiT-python/mhkit/dolfyn/io/rdi.py:75, in read_rdi(filename, userdata, nens, debug, vmdas_search, winriver, search_num, **kwargs)
     66 # Reads into a dictionary of dictionaries using netcdf naming conventions
     67 # Should be easier to debug
     68 rdr = _RDIReader(
     69     filename,
     70     debug=debug,
   (...)
     73     search_num=search_num,
     74 )
---> 75 datNB, datBB = rdr.load_data(nens=nens)
     77 dats = [dat for dat in [datNB, datBB] if dat is not None]
     79 # Read in userdata

File [...]/MHKiT-python/mhkit/dolfyn/io/rdi.py:355, in _RDIReader.load_data(self, nens)
    353 for dat, cfg in zip(datl, cfgl):
    354     dat, cfg = self.cleanup(dat, cfg)
--> 355     dat = self.finalize(dat, cfg)
    356     if "vel_bt" in dat["data_vars"]:
    357         cfg["rotate_vars"].append("vel_bt")

File [...]/MHKiT-python/mhkit/dolfyn/io/rdi.py:1047, in _RDIReader.finalize(self, dat, cfg)
   1045     cfg["fs"] = round(1 / np.median(np.diff(dat["coords"]["time"])), 2)
   1046 else:
-> 1047     cfg["fs"] = 1 / (cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"])
   1049 # Save configuration data as attributes
   1050 dat["attrs"] = cfg

ZeroDivisionError: float division by zero
```

The data is collected in burst mode and with both water column data and bottom track. The reason for the division-by-zero error is that `cfg["fs"]` is 0.

If I comment out line 1047 in mhkit/dolfyn/io/rdi.py (or use line 1045 instead) the function works and the file is read correctly.

### Desktop:
- MHKiT Version: 0.9.0

simmsa:
## Summary

Fix #408: ZeroDivisionError when reading RDI burst mode data files where `sec_between_ping_groups=0` causes division by zero in sampling frequency calculation.

## Changes

- Added check for `sec_between_ping_groups == 0` before calculation of `fs`.

  - When true, set `fs` to `NaN` with descriptive warning.

- Added test using `unittest.mock.patch` to verify fix behavior with burst mode configuration.

- Fixed pytest-asyncio deprecation warning in test configuration.

## Details

Per #408 RDI instruments in continuous burst mode set `sec_between_ping_groups=0` for maximum ping rate, causing ZeroDivisionError in:

```python
cfg["fs"] = 1 / (cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"])
```

The fix detects this condition and sets fs=NaN since burst mode operation with variable timing prevents reliable sampling rate determination using standard methods.

## Testing

Patch RDIReader.finalize to set `sec_between_ping_groups` to `0` and verify the fix prevents the `ZeroDivisionError`, sets `fs` to `np.nan` and warns the user.

### DIFF
--- a/mhkit/dolfyn/io/rdi.py
+++ b/mhkit/dolfyn/io/rdi.py
@@ -124,7 +124,7 @@ def read_rdi(
 
     if len(dss) == 2:
         warnings.warn(
-            "\nTwo profiling configurations retrieved from file" "\nReturning first."
+            "\nTwo profiling configurations retrieved from file\nReturning first."
         )
 
     # Close handler
--- a/mhkit/dolfyn/io/rdi.py
+++ b/mhkit/dolfyn/io/rdi.py
@@ -1053,7 +1053,9 @@ class _RDIReader:
                 )
                 cfg["fs"] = np.nan
             else:
-                cfg["fs"] = 1 / (cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"])
+                cfg["fs"] = 1 / (
+                    cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"]
+                )
 
         # Save configuration data as attributes
         dat["attrs"] = cfg
--- a/mhkit/dolfyn/io/rdi.py
+++ b/mhkit/dolfyn/io/rdi.py
@@ -1044,7 +1044,16 @@ class _RDIReader:
         ) or ("sentinelv" in cfg["inst_model"].lower()):
             cfg["fs"] = round(1 / np.median(np.diff(dat["coords"]["time"])), 2)
         else:
-            cfg["fs"] = 1 / (cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"])
+            # Handle burst mode where sec_between_ping_groups or pings_per_ensemble may be 0
+            if cfg["sec_between_ping_groups"] == 0 or cfg["pings_per_ensemble"] == 0:
+                warnings.warn(
+                    "mhkit.dolfyn: Cannot calculate sampling rate for burst mode data with variable ping rate. "
+                    f"Setting fs to NaN. sec_between_ping_groups={cfg['sec_between_ping_groups']}, "
+                    f"pings_per_ensemble={cfg['pings_per_ensemble']}."
+                )
+                cfg["fs"] = np.nan
+            else:
+                cfg["fs"] = 1 / (cfg["sec_between_ping_groups"] * cfg["pings_per_ensemble"])
 
         # Save configuration data as attributes
         dat["attrs"] = cfg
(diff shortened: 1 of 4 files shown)


simmsa:
@jmcvey3 thanks for the feedback. The median solution seems like a good compromise for this specific situation. 

I think this is ready to go, is there anything else needed in this PR?

jmcvey3:
> @jmcvey3 thanks for the feedback. The median solution seems like a good compromise for this specific situation.
> 
> I think this is ready to go, is there anything else needed in this PR?

Let me check to see if any of our current test files actually have this case. I can't remember, but the SentinelV test files I added might.

mhkit/dolfyn/io/rdi.py
```diff
@@ -1038,12 +1038,27 @@ def finalize(self, dat, cfg):
...
 
         # Need to figure out how to differentiate burst mode from averaging mode
-        if (
-            ("source_program" in cfg)
-            and (cfg["source_program"].lower() in ["vmdas", "winriver", "winriver2"])
-        ) or ("sentinelv" in cfg["inst_model"].lower()):
-            cfg["fs"] = round(1 / np.median(np.diff(dat["coords"]["time"])), 2)
+        calculate_sample_rate_from_time_diff = (
+            cfg.get("source_program", "").lower() in ["vmdas", "winriver", "winriver2"]
+            or "sentinelv" in cfg["inst_model"].lower()
```
jmcvey3, line 1043: I couldn't pull your branch, but please remove this line 1043. It's no longer necessary and your solution is cleaner.

jmcvey3 (changes requested):

mhkit/dolfyn/io/rdi.py
simmsa: I added you as a collaborator, you should be able to push changes now. The latest comments line 1043, are you indicating that line should be removed entirely? I can happily remove that line, I just want to make sure I am doing the right thing.

jmcvey3: Yes, you can remove all of `or "sentinelv" in cfg["inst_model"].lower()`. The tests should still pass

simmsa: Sounds good, removed that line.

jmcvey3 (approved):

PR #422, opened 2025-10-15
Fix failing pip tests

simmsa:
Status: Ready

This is a fix for failing pip tests in main.yml:

https://github.com/MHKiT-Software/MHKiT-Python/actions/runs/18511807517/job/52754343805?pr=421
https://github.com/MHKiT-Software/MHKiT-Python/actions/runs/18511807517/job/52754343883?pr=421

There are lots of errors in both of these tests related to h5 and netcdf4. One fix for this is installing the system h5 and nc binaries and having pip rebuild against these binaries

PR #423, opened 2025-10-16
Fix pip module tests

simmsa:
Status: Ready

This fixes test failures related to module loading using `pip install "mhkit[<module>]"` syntax.

There are some modules that have interdependencies with other modules, i.e., loads module needs wave module code that are causing test failures.

The root cause of test failures is specific packages that are not specified in pyproject.toml for a specific module, but are called when the tests are run.

The fix is to add lazy loading for the `wave.io` module and reconfigure the modules in pyproject.toml to capture the a more correct dependency list derived from running the tests.

Also the command that runs these tests was changed from `pip install -e "mhkit[<module>]" .` to `pip install -e ".[<module>]"` because the tests fail with `-e "mhkit"` but succeed with `-e "."`. I suspect that this changes the way mhkit is installed, but I don't know what the specific differences are.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,10 +10,6 @@ on:
       - main
       - develop
 
-concurrency:
-  group: ${{ github.event.pull_request.number }}  # Use the pull request number as the concurrency group
-  cancel-in-progress: true # Cancel any in-progress jobs for the same group. Prevents duplicate workflows on PRs from develop into main or from main into develop.
-
 jobs:
   set-os:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


simmsa:
@akeeste, this is good to go after the test-optional-pip-depencencies tests pass.

It may make sense to merge the 3 open PR's into develop and then merge that into main, but I will leave that up to you. If you go this route, LMK if there are merge conflicts, and I can fix them in each PR individually.

akeeste (commented):
@simmsa the previous install command is likely failing because it was using the version on pypi which is currently broken. Your change updates it to install the local github clone which contains the entire source. I think we should be using the git clone as you updated to since we do that everywhere else.

To my understanding, the changes to pyproject.toml will fix the pip-optional-dependency tests by themselves. The lazy loading of `mhkit.wave.io` is for a different convenience correct? 

I went forward with changes into main. Let's stick there until we get all of our bugfixes in and then we'll fast forward develop

simmsa:
> Also can you clarify the forced install of netcdf and hdf5 except for river, power, utils, loads? If those modules don't depend on netcdf and hdf5, which were causing issues that you solved in #422, then we can simplify their tests by treating them identically to the other tests right?

Because the module code uses pip to install it will have the same "limitations" as other pip installs and we need to perform the same steps to add the h5 and nc binaries and then rebuild against them. Each of the tests are "independent" pip installs so we need individually install the necessary dependencies. We exclude the modules where nc/h5 files are not used which you listed above.

PR #424, opened 2025-10-17
Actions: Remove --no-index from Wheel test build command

simmsa:
This is a fix for:

https://github.com/MHKiT-Software/MHKiT-Python/actions/runs/18594536347/job/53017269068#step:6:14

```
Run pip install 'mhkit[all]' --no-index --find-links dist/
Looking in links: dist/
Processing ./dist/mhkit-1.0.1-py3-none-any.whl
INFO: pip is looking at multiple versions of mhkit[all] to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement numpy>=2.0.0 (from mhkit[all]) (from versions: none)
ERROR: No matching distribution found for numpy>=2.0.0
Error: Process completed with exit code 1.
```

This fix removes `--no-index` which caused pip to only use local packages, which causes the above failure for any/all external packages/modules.

### DIFF
--- a/.github/workflows/test-wheel-build.yml
+++ b/.github/workflows/test-wheel-build.yml
@@ -1,0 +1,43 @@
+# Tests that the package can be built and installed from a wheel
+# Runs on PRs to catch build issues before merging to main
+# Does NOT publish to PyPI
+
+name: Test PIP Package Build
+
+on:
+  pull_request:
+    branches:
+      - main
+      - develop
+
+jobs:
+  test-build:
+    name: Test wheel build
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


simmsa:
@akeeste, this is good to go. This should solve the above test failure. This adds a new action: https://github.com/MHKiT-Software/MHKiT-Python/actions/runs/18597697846/job/53028120592 to verify the wheel build in develop. We may not need to keep this, but I wanted to verify that the pip command changes worked before we run the pypi.yml jobs.

akeeste (approved):
This looks good and will prevent future build failures from getting on pypi. The relevant tests are passing.

PR #426, opened 2025-10-17
Fix wheel build test

akeeste:
This PR 
- changes the pypi.yml action to use python 3.10 since 3.8 was breaking on the runner for unclear reasons
- updates the actions versions used for git checkout and setup-python too to those used successfully in [tsdat](https://github.com/tsdat/tsdat/blob/main/.github/workflows/pypi-release.yml)


simmsa (approved):
Looks good to me.

PR #429, opened 2025-11-24
D3D Grid Convergence example

browniea:
Update d3d capabilities to include Grid Convergence Index (GCI) calculation and example notebook


browniea:
I'm still waiting on more data to finish the example notebook and test the new GCI function

browniea:
@akeeste I'm thinking about removing everything about transects from the image and intro and saving it for another example notebook comparing the transects.

browniea:
- [ ] Convert example to xarray
- [ ] at the 16m and 32m grids to grid convergence

akeeste:
@browniea please review the changes here to see if there are any you want to keep, and then close. We can continue discussing the grid convergence notebook offline until it is finalized

PR #430, opened 2025-12-08
Updates for DOLfYN: improved handling of "averaged" profiles, ADCP waves example, critical PSD bugfix

jmcvey3:
Collection of bugfixes and documentation for Nortek Signature ADCPs. 
I received a dual-profile datafile from a Nortek Signature250 deployed at PacWave that was collecting both water velocity and wave measurements, and I made some updates to the codebase when things would fail working through the standard ADCP workflow. 

DOLfYN is currently set up to return individual profiles as individual datasets, so there is one dataset containing the wave-relevant variables, and one containing the water velocity variables. Both datasets contain water velocity information, so to differentiate between the two, the "_avg" tag is added to the dataset containing what Nortek calls the "averaging" profile. However, this dataset no longer comes out of the box "bin-averaged"; individual pings from each duty cycle are now saved in the output file.

I've been making updates to dolfyn so that the codebase will recognize "_avg" variables and default to them if untagged variables do not exist. The following updates are the latest:
- The "U_mag" and "U_dir" shortcuts will using "vel_avg" to calculate speed and direction if "vel" does not exist in the dataset.
- The shear functions in the ADCP turbulence API (`dudz`, `dvdz`, etc) can now utilize "vel_avg" if given as an input.
- `calc_declination` will no longer complain if you try to update the magnetic declination in the Nortek-created bin-averaged binary file ("<filename>_avgd.ad2cp")

Two, I created an example notebook showing how to calculate wave statistics using DOLfYN's FFT tools. We've gotten questions about this in the past, and now that I have a good dataset, this is a good time to document this.

Three, I found a bug in the PSD functions where individual FFTs get a 50% overlap not once, but twice. The core FFT function applies a 50% overlap using a series of "for loops" (the more robust method), while the input FFT function `cpsd` was adding overlap via the "npad" input to the `reshape` function. This latter method pads the first and last FFT with a lot of zeros, which in turn corrupts the first and last spectrum of a timeseries. It appears the latter method was written first and improved upon via the second method and should have been removed. Removing it fixes said bug.


jmcvey3:
@akeeste @simmsa This is at a good point for review. I'm happy to take feedback on the example notebook I added here

mhkit/dolfyn/velocity.py
```diff
@@ -304,7 +304,10 @@ def u(
         - earth:     east
         - principal: streamwise
         """
-        return self.ds["vel"][0].drop_vars("dir")
```
akeeste, line 307: In your first comment you mention defaulting to average quantities if they exist. This will default to returning non avg velocity components. Is this intended?

akeeste (approved):
Thanks @jmcvey3, great example notebook! I found it very clear and easy to walk through.

mhkit/dolfyn/velocity.py
jmcvey3: Oops, no, I edited the first comment to say "will return "avg" variables if untagged variables do not exist

PR #431, opened 2025-12-10
Branding: Update NLR branding

simmsa:
Update documentation to use National Laboratory of the Rockies per: https://www.energy.gov/eere/articles/energy-department-renames-nrel-national-lab-rockies

At this point in time <https://nrel.gov> is still active and <https://nlr.gov> does not work, so we are keeping the nrel.gov links the same.

Also the HSDS api endpoints:

`/nrel/US_wave/virtual_buoy/{region}/{region}_virtual_buoy_{year}.h5`
`/nrel/wtk/{region.lower()}/{region}_*.h5`
`/nlr/wtk/{region.lower()}-5min/{region}_*.h5`

In

```
mhkit/wave/io/hindcast/hindcast.py
mhkit/wave/io/hindcast/wind_toolkit.py
```

Are not updated yet, but may change in the future.


akeeste (approved):
Thanks Andrew, approved and merging

PR #421, opened 2025-10-14
Fix pip module package building

simmsa:
Status: Ready

The v1.0.0 release on PyPI is missing all submodules (wave, tidal, dolfyn, etc.) when installed via pip and downstream via conda.

## Reproduction

```bash
pip install "mhkit[all]"
```

```bash
python
Python 3.11.12 (main, Apr  8 2025, 14:15:29) [Clang 15.0.0 (clang-1500.1.0.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mhkit
>>> mhkit.__version__
'v1.0.0'
>>> from mhkit import river
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'river' from 'mhkit' (/opt/homebrew/lib/python3.11/site-packages/mhkit/__init__.py)
```

### Installed Package Contents

Inspecting the installed package shows two files (run `python -m site` to find `site-packages`
path):

```bash
ls /opt/homebrew/lib/python3.11/site-packages/mhkit
__init__.py __pycache__ warnings.py
```

All submodules (wave, river, tidal, dolfyn, power, loads, mooring, acoustics, qc, utils) are missing.

### Reproduction

Running the same build command used by GitHub Actions:

```bash
python -m build --wheel --outdir dist/ .
```

- `python -m build`: Runs Python's `build` module (PEP 517 build frontend) to build the package
- `--wheel`: Build only a wheel distribution (skips source distribution)
- `--outdir dist/`: Output directory for built files
- `.`: Build the package in the current directory (where `pyproject.toml` is)

Build output shows two files being copied:

```
running build_py
creating build/lib/mhkit
copying mhkit/warnings.py -> build/lib/mhkit
copying mhkit/__init__.py -> build/lib/mhkit
```

Wheel contents:

```
adding 'mhkit/__init__.py'
adding 'mhkit/warnings.py'
```

## Probable Root Cause

In `pyproject.toml`, this configuration tells setuptools to only include the top-level `mhkit` package, not any subpackages per: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration

```toml
[tool.setuptools]
packages = ["mhkit"]
```

## Proposed Fix

Replace the explicit package list with automatic package discovery using setuptools' `find` directive.

[Setuptools Package Discovery Documentation](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages)

```toml
[...]

Issue #303, opened 2024-04-09
hdf5 headers missing


nikg321:
### To Reproduce:
Provide a minimal working example that reproduces the error. 

**Minimal working example:** A minimal working example is a collection of source code and other data files which allow a bug or problem to be demonstrated and reproduced. The important feature of a minimal working example is that **it is as small and as simple as possible**, such that it is just sufficient to demonstrate the problem, but without any additional complexity or dependencies which will make resolution harder.

### Desktop (please complete the following information):
 - MHKiT Version: [e.g. 0.4.0]

ssolson:
Hello @nikg321 and thank you for your interest in MHKiT.

Without the additional context requested by the issue template I cannot help you. 

Please provide additional context about how the issue is occurring so that I can help you.

ssolson:
@nikg321 you said when you try to install the package you get a netCDF 4 error in your duplicate issue #305.

How are you trying to install MHKiT?

What operating system do you use?

What version of Python do you use?

nikg321:
I try using the pip install mhkit and I hot the same error by trying git
git clone,windows 64 the python version is 3.11.7 I think

Στις Τετ 10 Απρ 2024, 9:10 μ.μ. ο χρήστης ssolson ***@***.***>
έγραψε:

ssolson:
If you have cloned the repo can you try checking out the develop branch and seeing it installs correctly?

```
cd MHKiT-Python
git checkout develop
pip install -e .
```

nikg321:
Ok, I'll check it out.

Στις Τετ 10 Απρ 2024 στις 10:15 μ.μ., ο/η ssolson ***@***.***>
έγραψε:

nikg321:
Successfully installed NREL-rex-0.2.86 aiobotocore-2.5.0 aiofiles-23.2.1
aiohttp-3.8.5 aiohttp-cors-0.7.0 aioitertools-0.11.0 aiosignal-1.3.1
async-timeout-4.0.3 azure-core-1.30.1 azure-storage-blob-12.19.1
beautifulsoup4-4.12.3 botocore-1.29.76 bottleneck-1.3.8 fatpack-0.7.7
frozenlist-1.4.1 h5pyd-0.18.0 hsds-0.8.4 isodate-0.6.1 jmespath-1.0.1
joblib-1.4.0 lxml-5.2.1 mhkit-0.7.0 multidict-6.0.5 numcodecs-0.12.1
[...]

Issue #177, opened 2022-07-12
NETCDF 1.6.0 Breaking Tests


ssolson:
The latest version of NETCDF has breaking changes for MHKiT's tests during the installation process of the tests.

As an immediate fix, I will require the previous version of NetCDF in the Develop branch to complete the current PR to update the Development to the master branch for PR #176.

I will follow up on the error below in a future PR.

```
Downloading netCDF4-1.6.0.tar.gz (774 kB)
     -------------------------------------- 774.2/774.2 kB 8.2 MB/s eta 0:00:00
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  python setup.py egg_info did not run successfully.
  exit code: 1
  [29 lines of output]
  reading from setup.cfg...
      HDF5_DIR environment variable not set, checking some standard locations ..
  checking C:\Miniconda\envs\test\include ...
  hdf5 headers not found in C:\Miniconda\envs\test\include
  checking C:\Miniconda\envs\test\Library\include ...
  hdf5 headers not found in C:\Miniconda\envs\test\Library\include
  checking C:\Users\runneradmin\include ...
  hdf5 headers not found in C:\Users\runneradmin\include
  checking /usr/local\include ...
  hdf5 headers not found in /usr/local\include
  checking /sw\include ...
  hdf5 headers not found in /sw\include
  checking /opt\include ...
  hdf5 headers not found in /opt\include
  checking /opt/local\include ...
  hdf5 headers not found in /opt/local\include
  checking /opt/homebrew\include ...
  hdf5 headers not found in /opt/homebrew\include
  checking /usr\include ...
  hdf5 headers not found in /usr\include
  Traceback (most recent call last):
    File "<string>", line 36, in <module>
    File "<pip-setuptools-caller>", line 34, in <module>
    File "C:\Users\RUNNER~1\AppData\Local\Temp\pip-install-1ff035lt\netcdf4_b33f62e6324c4054b6b47aa5a10d45a5\setup.py", line 444, in <module>
      _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs)
    File "C:\Users\RUNNER~1\AppData\Local\Temp\pip-install-1ff035lt\netcdf4_b33f62e6324c4054b6b47aa5a10d45a5\setup.py", line 385, in _populate_hdf5_info
Error:       raise ValueError('did not find HDF5 headers')
[...]

PR #435, opened 2026-02-19
Package and testing updates and fixes

jmcvey3:
A general PR to cover tests that have been breaking recently to python and dependency updates. Also contains cleanup of the package in general
1. Makes the caching code in utils/cache.py less opaque. To properly test cache code, one needs to delete the .cache/mhkit folder, wherever that is stored on your local
2. Cleans up a new cache bug in wave/io/cdip.py
3. Packaging updates:
    - Part of our CI troubles may be that dependencies are double-installed via the conda environment.yaml file and the pyproject.toml file. Best practice is to use the environment file for conda-specific packages and keep dependencies in the .toml file. Dependencies will come from conda if the mhkit package is installed via conda instead of pip.
    - One possible issue may come from the hdf5-related packages. I'm aware there can an issue with hdf5 C library filepaths if they are installed partly via conda and partly via pip.
    - Updates github actions CI versions


jmcvey3:
And the wind-hindcast-cache tests have a dtype error. Something was changed and now the data returns as float64 instead of float32

simmsa:
@jmcvey3 thank you for breaking this out into a separate PR. 

There is a non obvious "test" here where we are verifying that the conda-forge install works. The original environment.yml is the same as this: https://github.com/conda-forge/mhkit-feedstock/blob/main/recipe/meta.yaml which is what conda-forge uses to build the conda version of mhkit.

Also important is that conda-forge uses this line:

```
{{ PYTHON }} -m pip install .[all] -vv --no-deps --no-build-isolation
```

to build the conda forge package. 

TBH this is a bit of python packing magic that I don't fully understand. In lieu of understanding, the tests build mhkit using a similar workflow as conda forge and basically "verify" that conda-forge will work, or not work. I think we should maintain this pattern and restore the environment.yml to be the same as the conda forge version. 
 
Having to define the same versions in two different files, pyproject.toml and environment.yml is really annoying, and maintaining a 3rd file in another repo even more annoying. I haven't seen any solutions for this in python, but maybe something better exists.

I think it is possible to autogen environment.yml from pyproject.toml to keep versions the same, so maybe that helps here. Or some other tool altogether.

Or maybe we need some other packaging solution that handles H5/HDF5 binaries better.

I'd be happy to try to pick this up, and try to get the tests working. Let us know what makes sense here.

jmcvey3:
@simmsa Thank you for the background here, this is starting to ring a bell

So to be clear, the only reason the extra dependencies are in environment.yml is to test and make sure the conda-forge meta.yaml will function properly? 

If so, should we move the original environment.yml file into something like "mhkit/tests/conda-forge/enviroment.yml" and leave a barebones environment.yml file in the main repo directory? That way we can refer to the former in the "conda-ubunut-latest" tests and the latter in the "pip-ubuntu-latest" tests and separate them properly?

pyproject.toml
```diff
@@ -54,8 +56,6 @@ river = [
 ]
 
 dolfyn = [
```
jmcvey3, line 60: I also saw that the h5py libraries are set here, but dolfyn doesn't actually require any of the h5-related libraries. These might belong under the wave module dependencies

simmsa: @jmcvey3 every netcdf file I have worked with is a h5 file: https://docs.unidata.ucar.edu/netcdf-c/4.9.3/interoperability_hdf5.html

Historically for mhkit it has been beneficial to control the Hdf5 library versions because it is typical for different libraries to use different versions, see #351, Issue #303, Issue #177.

simmsa:
> @simmsa Thank you for the background here, this is starting to ring a bell
> 
> So to be clear, the only reason the extra dependencies are in environment.yml is to test and make sure the conda-forge meta.yaml will function properly?
> 
> If so, should we move the original environment.yml file into something like "mhkit/tests/conda-forge/enviroment.yml" and leave a barebones environment.yml file in the main repo directory? That way we can refer to the former in the "conda-ubunut-latest" tests and the latter in the "pip-ubuntu-latest" tests and separate them properly?

Def not the only reason. environment.yml in the root allows developers to use conda to install mhkit packages. It seems valuable to support multiple installation methods for developers, which is why we define the dependencies twice.

pyproject.toml
jmcvey3: @simmsa I'm just commenting that they were listed under the dolfyn module and don't need to be. I learned when refactoring dolfyn that netCDF4 could incorporate h5 capabilities if needed, but it doesn't have to. The python h5 packages are bulky, so I wrote dolfyn to write netCDF4 files to only need the python netCDF4 dependencies.

jmcvey3:
> > @simmsa Thank you for the background here, this is starting to ring a bell
> > So to be clear, the only reason the extra dependencies are in environment.yml is to test and make sure the conda-forge meta.yaml will function properly?
> > If so, should we move the original environment.yml file into something like "mhkit/tests/conda-forge/enviroment.yml" and leave a barebones environment.yml file in the main repo directory? That way we can refer to the former in the "conda-ubunut-latest" tests and the latter in the "pip-ubuntu-latest" tests and separate them properly?
> 
> Def not the only reason. environment.yml in the root allows developers to use conda to install mhkit packages. It seems valuable to support multiple installation methods for developers, which is why we define the dependencies twice.

I don't understand - you can still use conda with a smaller environment.yml file. If you're downloading from conda-forge, then you'll get the packages from conda-forge. There's nothing different between packages downloaded from conda-forge and those downloaded from pip, assuming the versions are the same.

simmsa:
@jmcvey3 one other thing that you are may be fighting is the optional dependencies settings:

https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies

We set those to allow pip install "mhkit[<module>]" which allows module level installs, but I don't think any of us realized the implications of module level installs and the complications. 

I pushed a couple of fixes: #421, #422, #423 and #424 when we figured out that `conda install -c conda-forge mhkit` effectively installed nothing, but there was some nuance here that may be causing some of the problems now: https://github.com/MHKiT-Software/MHKiT-Python/blob/6f806c86f1c74a1d4e4765418bd0d7731fa0b8a6/pyproject.toml#L129-L142

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-namespace-packages

Do you think the above failures have anything to do with the module changes?

jmcvey3:
@simmsa I don't think I'm fighting the optional dependency installs; those seem to be working fine. I do remember having an issue with the package initialization process running into errors trying to import optional packages, but I think that was resolved.

pyproject.toml
akeeste: I support moving them to wave dependencies then. Part of the original motivation of the optional install set-up was to allow for small installs and keep the h5 packages in Dolfyn will force the module to be larger than necessary

akeeste (commented):
A couple clarifying questions on this PR:
- why the `--no-deps` change in the Actions workflows? What problem was this flag causing?
- `numexpr` and `pyarrow` are added to the `pyproject.toml` but I don't see them imported anywhere. Where are these two packages used?

And to confirm my understanding of the `environment.yml` files:
So now the all-inclusive `tests/conda-forge/environment.yml` serves the purpose of testing a pure-conda installation of MHKiT (plus the local GitHub version of MHKiT) used on a given PR. While `mhkit/environment.yml` is used in the pip installation of MHKiT and only installs python, pip, and hdf5 with conda before setting up the rest of the environment with pip

Lastly, my `test_cdip.py` is not working locally. I need to evaluate this further to see what is going on.

jmcvey3:
> A couple clarifying questions on this PR:
> 
> * why the `--no-deps` change in the Actions workflows? What problem was this flag causing?
> * `numexpr` and `pyarrow` are added to the `pyproject.toml` but I don't see them imported anywhere. Where are these two packages used?
> 
> And to confirm my understanding of the `environment.yml` files: So now the all-inclusive `tests/conda-forge/environment.yml` serves the purpose of testing a pure-conda installation of MHKiT (plus the local GitHub version of MHKiT) used on a given PR. While `mhkit/environment.yml` is used in the pip installation of MHKiT and only installs python, pip, and hdf5 with conda before setting up the rest of the environment with pip
> 
> Lastly, my `test_cdip.py` is not working locally. I need to evaluate this further to see what is going on.

For the "--no-deps" changes, the "pip-ubuntu-latest" tests were installing all dependencies via the conda environment file, rather than letting pip do the package dependency handling. This basically means conda installed all the packages, and we aren't testing if the versions that pip wants to install also function. Sometimes conda-approved versions are behind those on pip.

Good point, we should check if we still need to set "numexpr" and "pyarrow". They're dependencies for other packages, and I think we have them set because of some problematic package with a bad version release.

Yes, your understanding is correct. We can probably also test if we need hdf5 installed via conda if there is a user out there that wants to use venv or some such.

Yeah the "test_cdip.py" issues were what launched this whole escapade. Make sure you clear your mhkit cache on your local first. The first time you run the test, it writes the cache, and every time after you run the test, it reads the cache.

akeeste:
closing in favor of #436

PR #436, opened 2026-02-25
Package / Actions Cleanup - Continuation of #435

simmsa:
This is a continuation and refinement of #435

Highlights:

## From #435

* Actions version updates: 
  * checkout - v4 to v6
  * setup-python - v2 to v6
* cdip fixes
 
### Module Dependencies
  
* Pin pandas below version 3.0
  * There are some subtle bugs related to [underlying pandas changes](https://pandas.pydata.org/docs/user_guide/migration-3-strings.html#string-migration-guide) that we should address and possibly fix that are out of scope of this PR.
* Refactor dependencies to move xarray, netcdf, and h5 dependencies into `all`
  * This helps standardize the versions for all modules, and behind the scenes xarray typically requires both netcdf and h5.
  
### Conda and Conda-Forge Dependency Definitions

* Move conda forge environment file into environment-dev.yml and create new conda-forge-build job to test the conda-forge specific no deps build
  
### Linting  
       
* Add filter to only run `black` on files that have changed
  * Intent is to reduce noise in the tests. It is a non goal to have all files share the same formatting, we are just trying to enforce formatting on new or changed code
* Add Development Installation section to readme detailing pip, conda, and conda forge installation pathways

### Coveralls

* Pin coveralls action at v2
* Adds a switch at the top of main.yml to toggle reporting coveralls upload failures as test failures. Coveralls failures are of the 500 variant, ` Error: Bad response: 530 error code: 1016` per: https://status.coveralls.io/ and this implements the recommended `fail-on-error: false` fix

```
Update - Outage will be extended for at least several more hours. To avoid further disruption to your CI workflows, we recommend employing the fail-on-error: false input option available with all official coveralls integrations. See: https://docs.coveralls.io/integrations#official-integrations.
Feb 24, 2026 - 19:28 PST
```

We can incorporate this into #435 if that makes sense. My main goal here is to understand what changes are necessary in #435


simmsa:
@akeeste and @jmcvey3 this is ready.

Do any of you have a preference if we add this to #435 or keep this as a separate PR? Happy to do whatever makes sense.

There are lots of small test/action errors, warnings, and nits that this fixes. 

Also coveralls is still down: https://status.coveralls.io/. This adds a flag to ignore coveralls upload errors but that has the potential to lead to false positives (tests fail, but old passing test badges are shown to the user).

jmcvey3:
> @akeeste and @jmcvey3 this is ready.
> 
> Do any of you have a preference if we add this to #435 or keep this as a separate PR? Happy to do whatever makes sense.
> 
> There are lots of small test/action errors, warnings, and nits that this fixes.
> 
> Also coveralls is still down: https://status.coveralls.io/. This adds a flag to ignore coveralls upload errors but that has the potential to lead to false positives (tests fail, but old passing test badges are shown to the user).

It's fine by me to keep this separate and close 435. We want to open another PR to specifically address pandas 3.0?

.github/workflows/main.yml
```diff
@@ -10,39 +10,57 @@ on:
...
     steps:
       - id: set-matrix
         run: |
-          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "develop" ]]; then
-            echo "matrix_os=[ \"ubuntu-latest\"]" >> $GITHUB_OUTPUT
-          else
+          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "main" ]]; then
             echo "matrix_os=[\"windows-latest\", \"ubuntu-latest\", \"macos-latest\"]" >> $GITHUB_OUTPUT
+          else
+            echo "matrix_os=[\"ubuntu-latest\"]" >> $GITHUB_OUTPUT
```
akeeste: It's more expensive, but I prefer the previous set-up. The full test suite on main and dev serve as our primary flags for whether the tests are passing or not and don't typically hold us up--except at release time when we need to be especially robust in our testing anyways. Pushes to main/develop should also run the full test suite (all OS, always the hindcast, etc).

I'd focus decreasing test expense on PRs to develop as they are the primary bottleneck during active development. At the very least pushes to main to should run on all OS options.

simmsa:
@jmcvey3 

> It's fine by me to keep this separate and close 435. We want to open another PR to specifically address pandas 3.0?

Yes, I think we should open another PR for pandas 3.0. I was a little surprised that we needed to explicitly include `pyarrow` as a dep for mhkit. Seems like that should be a part of pandas requirements and not something MHKiT should have to manage.

simmsa:
Coveralls is operational now: https://status.coveralls.io/

image

I'm turning off the flag that ignores coveralls errors and this should pass and report to coveralls now...

akeeste:
@simmsa is this ready for review? I was in the midst of reviewing but see more commits coming in. Please convert to a draft if not ready yet.

simmsa:
> @simmsa is this ready for review? I was in the midst of reviewing but see more commits coming in. Please convert to a draft if not ready yet.

On the whole this is ready.

As a sanity check, the last push just verifies that changes made here correctly report to coveralls when it IS operational.

simmsa:
@akeeste this is good to go, apologies for the late push. Tests are passing with coveralls.

.github/workflows/main.yml
```diff
@@ -10,39 +10,57 @@ on:
...
           filters: |
             wave_io_hindcast:
               - 'mhkit/wave/io/hindcast/**'
               - 'mhkit/tests/wave/io/hindcast/**'
 
-      - id: hindcast-logic
+      - name: Determine if hindcast tests should run
+        id: hindcast-logic
+        if: github.event_name == 'pull_request'
         run: |
```
akeeste, line 66: Pushes also need to run the hindcast tests. See above comment. If we're getting false positives, then we should fix path-filter instead of limiting these tests to only PRs

.github/workflows/main.yml
```diff
@@ -238,7 +264,13 @@ jobs:
...
           conda activate mhkit-env
-          pip install -e . --no-deps
+          pip install -e ".[all,dev]"
+
+      - name: Download data from artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: data
+          path: ~/.cache/mhkit
```
akeeste: good catch here

environment-dev.yml
```diff
@@ -0,0 +1,28 @@
+name: mhkit-env
```
akeeste: suggested to change the environment name for clarify in one of the .yml files

README.md
```diff
@@ -82,6 +82,35 @@ To install all MHKiT dependencies use:
...
+pip install -e ".[all,dev]"
+```
+
+**Conda development** (minimal conda + pip resolves deps):
+
+```bash
+conda env create -f environment.yml
+conda activate mhkit-env
+pip install -e ".[all,dev]"
+```
```
akeeste, line 250: Is this really a good use case to support and test? What's the value in testing a combined pip/conda install? I recall that the various hdf5/h5 packages are problematic, but if a pure pip install works and a pure conda install works, where will this method be used?

Regardless if we remove one of the .yml files all together, I think one of them should be bumped to the mhkit/tests directory so the user install method is clear

akeeste (changes requested):

.github/workflows/main.yml
simmsa: Sounds good. I think that is fair, and makes sense. Happy to change this back.

.github/workflows/main.yml
simmsa: The terminology is a bit ambiguous here from git so let me make sure I understand the intent.

There are 3 situations I see (I'm likely missing some):

1. A developer is working on module code not related to hindcast
2. A developer is working on infra/actions/package code
3. A developer is working on hindcast code

For 1 I don't think we should ever run the hindcast tests. They are too prone to failure, even for merges into develop. They add "noise" to the tests.

For 2, I think they should run ideally once, and maybe this is after a merge into develop, but this is the hardest case to figure out the right thing to do. This is likely when we would figure out the tests are failing because of some external change and need to intervene, but "when" is the question here.

For 3. The tests need to run every time a change is pushed, but don't necessarily need to run on merge, but maybe a merge run doesn't hurt. 

@akeeste, what is your opinion here?

environment-dev.yml
simmsa: Good call, will fix that. `mhkit-dev-env` makes sense to me.

README.md
simmsa: This is @jmcvey3 recommended conda install method, and seems like a middle ground to me between a pip install -e ".[all]" and the pinned conda-forge dependency install. 

This supports a user or developer who wants to use conda to manage the environment and the h5 binaries but use pip to install everything else. 

I'd think of this as a "power user" or "conda developer" install basically.

I will probably migrate to using this locally to develop as sometimes conda-forge deps with other libraries and the solver can be really slow.

.github/workflows/main.yml
```diff
@@ -10,43 +10,72 @@ on:
...
     steps:
       - id: set-matrix
-        run: |
-          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "develop" ]]; then
-            echo "matrix_os=[ \"ubuntu-latest\"]" >> $GITHUB_OUTPUT
-          else
-            echo "matrix_os=[\"windows-latest\", \"ubuntu-latest\", \"macos-latest\"]" >> $GITHUB_OUTPUT
-          fi
+        run: echo "matrix_os=[\"windows-latest\", \"ubuntu-latest\", \"macos-latest\"]" >> $GITHUB_OUTPUT
+
```
akeeste: @simmsa Unless you think there's a strong advantage gained, I'd revert back to only ubuntu for PRs into develop and all OS elsewhere to keep our test suite fast when pushing many commits to open PRs
```suggestion
        run: |
          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "develop" ]]; then
            echo "matrix_os=[ \"ubuntu-latest\"]" >> $GITHUB_OUTPUT
          else
            echo "matrix_os=[\"windows-latest\", \"ubuntu-latest\", \"macos-latest\"]" >> $GITHUB_OUTPUT
          fi

File: .github/workflows/main.yml (hunk 2 above)
Comment (akeeste) [reply .github/workflows/main.yml:66]: All three of those categories fall under PRs but runs on merges fall into a different category to me. Also I believe the hindcast tests were updated to only ever actually hit the API on main. We'd need to double check, but if I recall correctly, cached data was always used for PRs and maybe merges into main. 

The hindcast tests should be very robust when we're not hitting the API, so perhaps we always run the hindcast tests, but only hit the API on merges into main or for PRs when the hindcast code is changed

File: README.md (hunk 5 above)
Comment (akeeste) [reply README.md:250]: That's fair that there is a use case here, but is it significant enough to warrant increasing test costs by 50%? Do the package versions installed by pure conda, pure pip and conda+pip differ enough to warrant the extra development cost?

File: .github/workflows/main.yml (hunk 2 above)
Comment (simmsa) [reply .github/workflows/main.yml:66]: @akeeste, sounds good. I reverted this to the original implementation because I don't think changes here added any value, and if anything changes made this more confusing.

File: .github/workflows/main.yml (hunk 6 above)
Comment (simmsa) [reply .github/workflows/main.yml]: Done.

File: README.md (hunk 5 above)
Comment (simmsa) [reply README.md:250]: @akeeste Fair question. If we offer a specific installation method it seems wise to test that installation method to verify it works. I don't think that this adds too much extra compute time, but happy to do whatever makes sense here.

Comment (jmcvey3) [reply README.md:250]: To clarify, for the full pip install, are we using venv for environment control? I'm well versed in conda now, so I use it to handle environments and python versioning, even if I don't install conda-specific packages.

If the full pip install is functional, then a.) yes I agree this test case is redundant, and b.) we can remove the h5 binaries from the basic conda environment.yml file.

File: .github/workflows/main.yml [hunk 7]
```diff
@@ -248,12 +278,80 @@ jobs:
...
+          python-version: ${{ matrix.python-version }}
+          use-only-tar-bz2: false
+          channels: ${{ env.CONDA_CHANNELS }}
+          conda-remove-defaults: ${{ env.CONDA_REMOVE_DEFAULTS }}
+
+      - name: Create MHKiT Conda environment
+        shell: bash -l {0}
+        run: |
+          conda env create -f environment-dev.yml
+          conda activate mhkit-dev-env
```
jmcvey3: So, this "env create" line (319) is actually redundant after the "Setup Miniconda" routine in line 304. You just need line 320 here is activate the environment.

jmcvey3: @simmsa @akeeste So actually yeah, we're testing the full pip install using a conda environment every time. If we want a pure pip install, then we need to use venv as the package manager.

README.md
simmsa: @jmcvey3, this seems reasonable to me. I will refactor accordingly.

.github/workflows/main.yml
simmsa: @jmcvey3 good call on venv over conda for pip install testing. This is definitely a gap in the documentation and tests. 

I will work to refactor accordingly including updating the README documentation to recommend venv for pip installs and updating the tests.

simmsa: > So, this "env create" line (319) is actually redundant after the "Setup Miniconda" routine in line 304. You just need line 320 here is activate the environment.

Good find here, also. Will fix...

README.md
```diff
@@ -33,55 +33,142 @@ See the [MHKiT documentation](https://mhkit-software.github.io/MHKiT) for more i
...
+For developers contributing to MHKiT, there are three development installation strategies after
+cloning the repository locally:
+
+Note: Use of a virtual environment is recommended to avoid dependency conflicts with other python.
+See Option 2 installation instructions for virtual environment setup.
+
+**Pip development** (no conda):
+
+```bash
+pip install -e ".[all,dev]"
```
jmcvey3, line 235: Just adding a note to create and activate a venv environment here

simmsa: Good call, will update accordingly. LMK if other areas don't make sense here either, or could be clarified.

.github/workflows/main.yml
simmsa: This is fixed is the latest version

README.md
simmsa: This is also fixed.

simmsa:
@jmcvey3, I removed the CDIP changes in this PR, and the tests now pass. This suggests the test failures you saw were related to the pandas 3.0 update. I plan to use your changes in the upcoming PR for pandas 3.0, but just wanted to give you a heads up.

Issue #427, opened 2025-11-11
Understanding spread data for directional wave spectrum


jtgrasb:
I am working with the [directional wave example](https://mhkit-software.github.io/MHKiT/directional_waves.html) to create a streamlined approach between MHKiT full directional waves and WEC-Sim's approach to creating and modeling the waves. I am also trying to stay consistent with the IEC standard:

<img width="473" height="359" alt="Image" src="https://github.com/user-attachments/assets/4c4f9175-1b0e-474e-bfa6-9b8f37cafeb0" />

In WEC-Sim, the spread is defined such that it sums to 1 across all directions. Thus, the energy is maintained between the omni-directional spectrum (indexed by each frequency) and the full directional spectrum (indexed by frequency and direction). In the directional wave example, the matrix defining the spread does not sum to 1 (instead it sums to 28.65 for each frequency). When defining the full directional wave spectrum, should the spread be scaled to 1? Or is there something I'm missing in how the spread and spectrum are defined by MHKiT?

jtgrasb:
After reviewing further, my mistake is that the *integral* of the directional spread from -pi to pi should be equal to 1, not the sum. The MHKiT formulation satisfies this condition. I do think the units for the spread may be off though. In the example, the spread spectrum units are listed as 1/Hz/deg, but I believe they should be 1/Hz/rad.

akeeste:
Thanks @jtgrasb apologies for not getting to this earlier. Was the units issue in the example just a matter of labeling or is there a factor incorrect?  Also can you refer me to the where you see radians input? 

[Here](https://www.ndbc.noaa.gov/faq/measdes.shtml) I see degrees used for the spectral wave direction and the directional_waves.ipynb defines the directions using degrees

jtgrasb:
The units issue is simply a matter of labeling of the dataarray and on the plot, not too important. I am referring to the directional wave spectrum ( $D(f,\theta)$ ), the units for which are not defined by NDBC as far as I can tell.

<img width="600" height="344" alt="Image" src="https://github.com/user-attachments/assets/9714cab5-f3f3-4a34-846a-8b7056c55350" />

The key part of the directional wave spectrum is that the integral from -pi to pi radians is equal to 1, implying that it is in units of 1/Hz/rad. From the IEC standard (and consistent with the values from MHKiT:

[...]

PR #437, opened 2026-03-19
NDBC Updates

jmcvey3:
Opening this to solve Issue #427 

I changed the source code to return degrees instead of radians, which was a simple fix. I also went ahead and updated the polar plots so that zero degrees is at the top of the figure, and positive degrees runs clockwise. We should update this in other polar plots in MHKiT too.

I also updated units to pass CF conventions, whereafter I noticed that our tests for this code check that the units are the same as what NDBC outputs. I'm not sure that's necessary. I also saw that we're not actually testing the value output from the source code, which is important.


akeeste (approved):
Thanks for getting to this @jmcvey3. Approved, no changes from me.

Resolves issue #427

akeeste:
Tests are failing for other reasons and as far as I can tell the changes to the calculations here wouldn't be caught by them anyways. Our NDBC test for  `create_directional_spectrum` and `create_spread_function` only check the size and units of the outputs but no numerical values

PR #448, opened 2026-06-16
Updated ADCP examples 

browniea:
Adding visualizations to match the [MHKiT_MATLAB ADCP example](https://github.com/MHKiT-Software/MHKiT-MATLAB/blob/master/examples/acoustics_example.mlx) and addressing several issues:  

Qhull Error https://github.com/MHKiT-Software/MHKiT-Python/issues/444

D3D coordinate systems https://github.com/MHKiT-Software/MHKiT-Python/issues/442

PR #452, opened 2026-07-01
Refactor DOLfYN PSD Scripts to use Scipy.Signal.Welch

jmcvey3:
Big refactor I'd been intending to finish for a while.

This refactor solves a DOLfYN structural problem between its binning architecture and the traditional sliding-window architecture of Welch's algorithm. Because DOLfYN was using slices of bins to compute each PSD segment, it was impossible to ensure data overlap between one bin and the next, meaning the industry-standard 50% overlap cannot be accomplished. This PR removes the bin structure from the PSD computation and ensures overlap using a "step" argument, which is also now built into the bin structure to properly average dimensions for the PSD.

In doing this, DOLfYN's custom PSD code, which was a custom implementation of Welch's algorithm without the ability to overlap FFT segments, has been replaced with scipy.signal.welch. This has one breaking change, in that NaN's are no longer tolerated in the PSD calculation. 

To avoid breaking changes, the DOLfYN code will default to an overlap of 0%. This ensures array shapes will remain the same. On the other hand, since it's new, the Acoustics module will default to 50%.

The second breaking change is that the PSD code now renames the input time dimension to "time_psd". In this way, if a different overlap is used, and the PSD is saved into dataset with an averaged "time" dimension, the PSD will not follow the xarray default and go to nan.

Finally, I removed the float32 datatypes in certain functions since sometimes we do need float64 precision. Removing custom code means fft.py no longer exists, and tools/misc.py was renamed to tools.py


jmcvey3:
Apologies for a mess of repeat commits from a previous PR, and I'll update the tests when I have another moment

mhkit/tests/dolfyn/test_analysis.py
```diff
@@ -179,13 +186,14 @@ def test_adcp_turbulence(self):
...
         slope_check = bnr.check_turbulence_cascade_slope(
-            tdat["psd"].mean("time_b5"), freq_range=[0.4, 4]
+            tdat["psd"].mean("time_psd"), freq_range=[0.4, 4]
         )
         # Check noise subtraction in psd function
         tdat["psd_noise"] = bnr.power_spectral_density(
             dat["vel_b5"].isel(range_b5=len(dat["range_b5"]) // 2),
             freq_units="Hz",
-            noise=0.01,
+            noise=0.0001,
```
simmsa, line 193: Previous test used 0.01. Is it intentional that this change uses 0.001, an extra decimal?

mhkit/tests/dolfyn/test_analysis.py
```diff
@@ -179,13 +179,14 @@ def test_adcp_turbulence(self):
...
+            pct_overlap=0.5,
```
akeeste, line 194: Just confirming if this is intended to be 0.5 or 0 like the other calls?

mhkit/dolfyn/binned.py
```diff
@@ -339,9 +306,11 @@ def variance(self, arr, axis=-1, n_bin=None):
         out : numpy.ndarray
         """
 
-        return np.nanvar(self.reshape(arr, n_bin=n_bin), axis=axis, dtype=np.float32)
+        return np.nanvar(
+            self.reshape(arr, step=step, n_bin=n_bin), axis=axis, dtype=np.float32
```
akeeste: `float32` to `float64`?

mhkit/dolfyn/binned.py
```diff
@@ -360,7 +332,9 @@ def standard_deviation(self, arr, axis=-1, n_bin=None):
         out : numpy.ndarray
         """
 
-        return np.nanstd(self.reshape(arr, n_bin=n_bin), axis=axis, dtype=np.float32)
+        return np.nanstd(
+            self.reshape(arr, step=step, n_bin=n_bin), axis=axis, dtype=np.float32
```
akeeste: `float32` to `float64`?

examples/adcp_example.ipynb
akeeste, line 1: There are a lot of changes to this file here and in #448 but I don't see any conflicts. Are these changes merged together to include the plots from the MATLAB notebook that @browniea added in #448?

akeeste (commented):
Thanks @jmcvey3 . A couple minor comments and one question around the ADCP_example.ipynb

mhkit/tests/dolfyn/test_analysis.py
jmcvey3: Yes, I wanted to make sure to have a test of the 50% overlap

mhkit/tests/dolfyn/test_analysis.py
```diff
@@ -89,23 +89,28 @@ def test_calc_func(self):
...
+            dat["vel"],
+            freq_units="rad",
+            window="hann",
+        )
 
         tdat["stress_detrend"] = bnr.reynolds_stress(dat["vel"])
         tdat["stress_demean"] = bnr.reynolds_stress(dat["vel"], detrend=False)
         tdat["csd"] = bnr.cross_spectral_density(
-            dat["vel"], freq_units="rad", window="hamming", n_fft_coh=10
+            dat["vel"], freq_units="rad", window="hann", n_fft_coh=10
```
simmsa, line 102: Is the change from hamming to hann window intentional? Mostly want to make sure this change is not a typo.

examples/adcp_example.ipynb
jmcvey3: Oh shoot no I think those got lost

jmcvey3:
Okay, @akeeste @simmsa, I believe this is completed.

examples/adcp_example.ipynb
jmcvey3: K, that notebook is updated based on 448

mhkit/dolfyn/binned.py
```diff
@@ -370,19 +340,22 @@ def _psd_base(
         noise=0,
         n_bin=None,
         n_fft=None,
-        step=None,
+        pct_overlap=0.5,
```
simmsa, line 343: Can we put in a check in this method and also the `_csd_base` method to guard against obvious out of spec user inputs to the `pct_overlap` parameter, i.e. 50.0. 

```
if pct_overlap < 0 or pct_overlap > 1:
    raise ValueError(f"pct_overlap must be between 0 and 1, got {pct_overlap}.")
``` 

The specification is spelled out in the docstring, but I think we should try to prevent any out of spec values from passing into scipy through this line:

```
noverlap=int(pct_overlap * n_fft)
``` 

If the user does input an out of spec value it is likely that `signal.welch` will produce an error if noverlap > nperseg:

https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/signal/_spectral_py.py#L913

But the error will likely be confusing. I think it would be more correct catch this on the MHKiT side.

mhkit/dolfyn/velocity.py
```diff
@@ -980,20 +997,22 @@ def power_spectral_density(
...
         fs : float (optional)
           The sample rate. Default is `binner.fs`
-        window : string or array
-          Specify the window function.
-          Options = 1, None, 'hann', 'hamm'. Default = 'hann'
+        window: str
+          Type of window to apply to each FFT segment.
+          Options ('boxcar', 'hann', 'hamming', 'blackman', 'bartlett').
+          Default: 'hann'.
         noise : numeric or array
```
simmsa, line 1005: Is it relevant to link to the available window functions in scipy here:

https://docs.scipy.org/doc/scipy/reference/signal.windows.html#module-scipy.signal.windows

Or are we only supporting a subset of windowing functions?

simmsa (approved):
@jmcvey3, overall this looks good. I approve.

I have a few comments but I think we can address them after the release.

akeeste (approved):

mhkit/dolfyn/velocity.py
jmcvey3: Whoa that's a lot. I just gave the most common examples, but we can link to that page

mhkit/dolfyn/binned.py
jmcvey3: Yes, I should do this

mhkit/tests/dolfyn/test_analysis.py
jmcvey3: It is yes, I'm trying to make sure all of the options are covered in testing.

mhkit/tests/dolfyn/test_analysis.py
jmcvey3: It's intentional to reduce the number of nans in the resulting values. Makes the test a little more robust

jmcvey3:
Alright @simmsa I mentioned `scipy.signal.window` for window options in the docstrings, and added a check for `pct_overlap`. I'll merge tomorrow if there's nothing else

simmsa:
> Alright @simmsa I mentioned `scipy.signal.window` for window options in the docstrings, and added a check for `pct_overlap`. I'll merge tomorrow if there's nothing else

Thank you! Nothing else from me.

PR #454, opened 2026-07-10
Experiment: Squash PR 436

simmsa:
#436 did not follow the convention to do a rebase merge commit. This PR is an attempt to squash all of the commits from 436 into a single commit.


simmsa:
@jmcvey3, this was my experiment to try to fix the extra commits from #436. 

I tried a few variants of "scary" git commands and was able to squash my commits into one commit, but this always alters the git history by changing the commit hashes. 

I'm going to not merge this out of an abundance of caution. In the future we (I) can definitely do a better job trying to catch these types of things earlier.

akeeste:
> @jmcvey3, this was my experiment to try to fix the extra commits from #436.
> 
> I tried a few variants of "scary" git commands and was able to squash my commits into one commit, but this always alters the git history by changing the commit hashes.
> 
> I'm going to not merge this out of an abundance of caution. In the future we (I) can definitely do a better job trying to catch these types of things earlier.

I agree with this conclusion. let's go forward with the existing history and ensure we squash on PRs into main/develop (but not for develop-->main or vice-versa).

Issue #458, opened 2026-07-21
Does mhkit work with the Nortek Aquadopp?


ddileonardo:
Does mhkit work with the Nortek Aquadopp? Specifically, I have data from the legacy (pre-2024) aquadopp. It is an ADCP, but it isn't mentioned in the documentation anywhere.

If mhkit does support aquadopps, which output file extension should I be trying to read?

jmcvey3:
@ddileonardo Looks like I forgot to update the reader docstring... It does support Aquadopps, but I've only had test data for the profiling versions. For those it's the `.prf` file, for the single point meters it is the `.aqd` filetype.

ddileonardo:
Thanks! I have a .prf file. The instrument was set to profile.

I am getting a KeyError when I try to use dolfyn.read. 

```python
from mhkit import dolfyn

data_path = r'GS-LMR_CH2ADP_020526.prf'
ds = dolfyn.read(data_path)
```

```python
KeyError                                  Traceback (most recent call last)
Cell In[3], line 2
      1 data_path = r'GS-LMR_CH2ADP_020526.prf'
----> 2 ds = dolfyn.read(data_path)

File c:\Users\ddileonardo\.conda\envs\qbay\Lib\site-packages\mhkit\dolfyn\io\api.py:114, in read(fname, userdata, nens, **kwargs)
    112     func_map = dict(RDI=read_rdi, nortek=read_nortek, signature=read_signature)
    113     func = func_map[file_type]
--> 114 return func(fname, userdata=userdata, nens=nens, **kwargs)

File c:\Users\ddileonardo\.conda\envs\qbay\Lib\site-packages\mhkit\dolfyn\io\nortek.py:60, in read_nortek(filename, userdata, debug, do_checksum, nens, **kwargs)
     51     logging.basicConfig(
     52         filename=str(logfile),
     53         filemode="w",
     54         level=logging.NOTSET,
     55         format="%(name)s - %(levelname)s - %(message)s",
     56     )
     58 userdata = base._find_userdata(filename, userdata)
---> 60 rdr = _NortekReader(filename, debug=debug, do_checksum=do_checksum, nens=nens)
     61 rdr.readfile()
     62 rdr.cleanup()

File c:\Users\ddileonardo\.conda\envs\qbay\Lib\site-packages\mhkit\dolfyn\io\nortek.py:230, in _NortekReader.__init__(self, fname, endian, debug, do_checksum, bufsize, nens)
[...]

PR #457, opened 2026-07-10
Develop: Merge main after v1.1.0 release

simmsa:
Update develop with main release commits

PR #459, opened 2026-08-05
Edge case for datafiles with odd number of bins

jmcvey3:
Quick fix for classic Nortek profiling instruments.

Sometimes but not always, an extra fill byte is included at the end of profiles with odd numbers of bins.

Solution for Issue #458

### DIFF
--- a/mhkit/__init__.py
+++ b/mhkit/__init__.py
@@ -11,7 +11,7 @@ from mhkit.warnings import configure_warnings
 
 configure_warnings()
 
-__version__ = "v1.1.0"
+__version__ = "v1.1.1"
 
 __copyright__ = """
 Copyright 2019, Alliance for Energy Innovation, LLC under the terms of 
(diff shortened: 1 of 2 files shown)


jmcvey3:
@simmsa @akeeste Can we add this as a hotfix for the latest release? Might need to go directly into main instead of develop

simmsa:
@jmcvey3 Yes, I think that seems reasonable to merge directly into main.

I will handle the merge and will merge main back into develop. Looks like I forgot to merge back #457 so I will use that to get develop in sync in main after merging.

simmsa:
@jmcvey3, pushed. This made more sense to go develop -> main.

At this time, v1.1.1 has not been pushed to pypi as actions are down, https://www.githubstatus.com/, and we use them to push to pypi. Hopefully this gets fixed and happens automatically but I will revisit incase we need to figure out an alternative solution.

simmsa:
@jmcvey3, the action issues were resolved and 1.1.1 is live: https://pypi.org/project/mhkit/1.1.1/

Issue #461, opened 2026-08-07
Wheel building issue for dependency package on Windows platforms


jmcvey3:
There is an error when building the wheels for aiohttp and bitshuffle on Windows - these are from the `NLR-rex` package dependence on `hsds`, and occurs when running `pip install mhkit[wave]`.

`Collecting bitshuffle>=0.5.2 (from hsds>=0.8.4->NLR-rex[hsds]>=0.5.0->mhkit[all])`
`Collecting aiohttp==3.9.4 (from hsds>=0.8.4->NLR-rex[hsds]>=0.5.0->mhkit[all])`

```bash
Building wheels for collected packages: aiohttp, bitshuffle
  Building wheel for aiohttp (pyproject.toml): started
  Building wheel for aiohttp (pyproject.toml): finished with status 'error'
  Building wheel for bitshuffle (pyproject.toml): started
  Building wheel for bitshuffle (pyproject.toml): finished with status 'error'
Failed to build aiohttp bitshuffle

Pip subprocess error:
  error: subprocess-exited-with-error
  
  × Building wheel for aiohttp (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [106 lines of output]
      *********************
      * Accelerated build *
      *********************
      C:\Users\mcve343\AppData\Local\Temp\pip-build-env-dnyynhe1\overlay\Lib\site-packages\setuptools\dist.py:765: SetuptoolsDeprecationWarning: Licen.
      !!
      
              ********************************************************************************
              Please consider removing the following classifiers in favor of a SPDX license expression:
      
              License :: OSI Approved :: Apache Software License
      
              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************
      
      !!
        self._finalize_license_expression()
      running bdist_wheel
      running build
      running build_py
      creating build\lib.win-amd64-cpython-313\aiohttp
      copying aiohttp\abc.py -> build\lib.win-amd64-cpython-313\aiohttp
      copying aiohttp\base_protocol.py -> build\lib.win-amd64-cpython-313\aiohttp
      copying aiohttp\client.py -> build\lib.win-amd64-cpython-313\aiohttp
      copying aiohttp\client_exceptions.py -> build\lib.win-amd64-cpython-313\aiohttp
      copying aiohttp\client_proto.py -> build\lib.win-amd64-cpython-313\aiohttp
[...]

PR #462, opened 2026-08-07
Dependency: Downgrade NLR-rex

simmsa:
On Windows, the installing mhkit 1.1.1 yields a failed instillation caused by the bitshuffle package as detailed in #461. The root cause appears to be the new `NLR-rex[hsds]` dependency in both pyproject.toml and environment-dev.yml

This reverts to an older version and removes the `hsds` extra which is not required.

Additionally `NLR-rex` is not available on conda-forge, while an older version of NREL-rex is available. An issue was submitted to NLR-rex: https://github.com/NatLabRockies/rex/issues/229, but our team does not have an eta on a conda-forge of NLR-rex.

This also adds tests for wheel builds on windows in an attempt to catch this build error during CI/CD.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -527,6 +527,30 @@ jobs:
         run: |
           python -m pytest -c .github/workflows/pytest.ini mhkit/tests/${{ matrix.module }}/
 
+  windows-wheel-check:
+    # Verifies each mhkit extras group can be installed on Windows from prebuilt wheels alone,
+    # which is intented to mimic a user without windows MSVC compilers installed
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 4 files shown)


simmsa:
# MHKiT v1.1.2: Dependency Improvements


## Bug Fixes

* Downgrade NLR-rex to address #461 and add conda-forge support.

simmsa:
@jmcvey3 this should fix #461. Could you pull this branch down and see if you are able to get this installed via pip?

simmsa:
> @jmcvey3 this should fix #461. Could you pull this branch down and see if you are able to get this installed via pip?

I think just doing:

`pip install -e ".[all]"` is the equivalent command for a local install.

jmcvey3 (approved):
This is working on my end.

